PR #4, opened 2019-12-23
Adds package est with first function FileExists

ForestEckhardt:
I have created a package called `est` that I would like to contain functions that help with verifying the state of files. To start I have added a FileExists function that checks to see whether or not a given file exist. I think that this would be a good place to also add functions like checking for symlinks, seeing it a file is executable or not, et cetera.

### DIFF
--- a/est/file_exists.go
+++ b/est/file_exists.go
@@ -1,0 +1,17 @@
+package est
+
+import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


ForestEckhardt:
After discussion it was decided that this package was unnecessary

PR #8, opened 2020-03-31
Added support for multiple build plans during detection.

dmikusa-pivotal:
[#171748623]

Co-authored-by: Timothy Hitchener <thitchener@pivotal.io>

### DIFF
--- a/detect.go
+++ b/detect.go
@@ -58,7 +58,7 @@ type BuildPlan struct {
 
 	// Or is a list of additional BuildPlans that may be selected by the
 	// lifecycle
-	Or []BuildPlan `toml:"or,omitepty"`
+	Or []BuildPlan `toml:"or,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


detect.go
```diff
@@ -55,6 +55,10 @@ type BuildPlan struct {
 	// Requires is a list of BuildPlanRequirements that are required by this
 	// buildpack.
 	Requires []BuildPlanRequirement `toml:"requires"`
+
+	// Or is a list of additional BuildPlans that may be selected by the
+	// lifecycle
+	Or []BuildPlan `toml:"or,omitepty"`
```
ryanmoran, line 61: This should be `omitempty`, right?

dmikusa-pivotal: Yes. Good catch.

dmikusa-pivotal: I updated the PR.

PR #9, opened 2020-04-06
Adds zip functionality to vacation

ForestEckhardt:
This PR adds zip functionality to vacation. As it stands currently I am unsure how to include symlinking so I excluded it for now.

### DIFF
--- a/vacation/vacation.go
+++ b/vacation/vacation.go
@@ -7,6 +7,7 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


ryanmoran:
@ForestEckhardt what are you unsure about for adding symlink support?

ForestEckhardt:
@ryanmoran I am unclear how to replicate the symlink on the other side when unzipping the file. It was unclear to me at the time of writing this how to use the go library to extract a functional symlink from a zipped artifact.

ryanmoran:
I've added support for symlinks. This [issue](https://github.com/golang/go/issues/3720) was helpful to understand what should happen.

PR #10, opened 2020-04-06
Adds strip-components option to tar archives

ForestEckhardt:
Purposed interface for strip-component option on for tar archives.

### DIFF
--- a/vacation/vacation.go
+++ b/vacation/vacation.go
@@ -7,20 +7,24 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


ryanmoran:
Why do we need this? I thought we were trying to get archives that didn't have these superfluous directories.

ForestEckhardt:
@ryanmoran I agree that this is true for the cnbs but I have been trying to use this package in other projects and the inability to do this when the tar command allows me is frustrating. It may also be beneficial to people using this library that do not have strong control over their artifacts. My goal with this interface was to keep the default behavior what we are expecting but allow for expansion.

PR #18, opened 2020-05-11
Adds readme

### LINKED ISSUE #12: Add a README
There are some pretty decent docs at https://godoc.org/github.com/cloudfoundry/packit, but this repo has no README. We should at least include a link to these docs and preferably much more.

ryanmoran:
Resolves #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # packit
 
+[![GoDoc](https://img.shields.io/badge/pkg.go.dev-doc-blue)](http://pkg.go.dev/.)
+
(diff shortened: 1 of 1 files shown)


ryanmoran:
@ForestEckhardt I'm trying to use [goreadme](https://github.com/posener/goreadme) to keep this README up to date. It is the one choosing pkg.go.dev, which is really the right place for docs. However, it uses proxy.golang.org to find packages. Since we've never released a paketo-buildpacks version of packit, there is no package with that path in the index and thus no docs to link to... yet. We need to merge #17 and then this should work.

PR #69, opened 2020-08-27
Uses CNB_BUILDPACK_DIR env variable to set CNBPath

### LINKED ISSUE #62: Use CNB_BUILDPACK_DIR env var to determine buildpack filepath
Packit already exposes the path to the buildpack on the filesystem via the `packit.DetectContext.CNBPath` and `packit.BuildContext.CNBPath` struct fields. As of Buildpack API v0.3, the lifecycle will include a `CNB_BUILDPACK_DIR` environment variable when invoking the `bin/detect` and `bin/build` binaries via [RFC 0036](https://github.com/buildpacks/rfcs/blob/main/text/0036-cnb-buildpack-directory-env-var.md).

Packit should use this environment variable to pass this information to the buildpack instead of its [own process](https://github.com/paketo-buildpacks/packit/blob/2a1fca2b76051da33ac96130e0e5bb09d01c2baa/detect.go#L107) of reverse-engineering the path. It should do this only in cases where the environment variable is available, and continue to perform the fall-back mechanism for cases where it is not. The value of the environment variable should still be made available via these same struct fields.

floragj:
The old strategy is used if it is not available

Implements Buildpack API version 0.3 (and resolves #62 )

Signed-off-by: Timothy Hitchener <thitchener@pivotal.io>

### DIFF
--- a/build.go
+++ b/build.go
@@ -183,8 +183,8 @@ func Build(f BuildFunc, options ...Option) {
 		return
 	}
 
-	cnbPath := os.Getenv("CNB_BUILDPACK_DIR")
-	if cnbPath == "" {
+	cnbPath, ok := os.LookupEnv("CNB_BUILDPACK_DIR")
+	if !ok {
 		cnbPath = filepath.Clean(strings.TrimSuffix(config.args[0], filepath.Join("bin", "build")))
 	}
 
--- a/build.go
+++ b/build.go
@@ -178,7 +178,10 @@ func Build(f BuildFunc, options ...Option) {
 		return
 	}
 
-	cnbPath := filepath.Clean(strings.TrimSuffix(config.args[0], filepath.Join("bin", "build")))
+	cnbPath := os.Getenv("CNB_BUILDPACK_DIR")
+	if cnbPath == "" {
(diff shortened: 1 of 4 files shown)


build.go
```diff
@@ -183,7 +183,10 @@ func Build(f BuildFunc, options ...Option) {
 		return
 	}
 
-	cnbPath := filepath.Clean(strings.TrimSuffix(config.args[0], filepath.Join("bin", "build")))
+	cnbPath := os.Getenv("CNB_BUILDPACK_DIR")
+	if cnbPath == "" {
+		cnbPath = filepath.Clean(strings.TrimSuffix(config.args[0], filepath.Join("bin", "build")))
+	}
```
ryanmoran, line 189: I think its a bit clearer to use `os.LookupEnv` here as it provides an API that tells you if the env var is not set, which is what we really care about, instead of switching on the contents of the env var. While it should never really be an empty string, I think using `os.LookupEnv` is a bit more intention revealing.

```suggestion
	cnbPath, ok := os.LookupEnv("CNB_BUILDPACK_DIR")
	if !ok {
		cnbPath = filepath.Clean(strings.TrimSuffix(config.args[0], filepath.Join("bin", "build")))
	}
```

PR #70, opened 2020-08-27
Process: support process specific env variables

### LINKED ISSUE #63: Extend packit.Process to include process-specific environment variables
The addition of [RFC 0042](https://github.com/buildpacks/rfcs/blob/main/text/0042-process-specific-env.md) to the Buildpack API means that processes defined in a `launch.toml` can also declare their own set of environment variables that are located at `<layer>/env.launch/<process-type>`.

Packit can easily support this through its API by including a new `Environment` field on the `Process` type. When this field is populated, the `packit.Build` function will ensure that the `Environment` contents are written to the environment directory for this process.

arjun024:
Though processes can exist independent of a layer, the spec only allows
buildpacks to implement process specific env vars via a launch layer.
So, as a workaround if any of the processes have env vars attached to
it, a launch layer `packit-process-launch` is created which houses the
`env.launch/<process>` directory from where the lifecycle reads these
process specific env vars

Signed-off-by: Arjun Sreedharan <asreedharan@vmware.com>

Resolves #63

### DIFF
--- a/build.go
+++ b/build.go
@@ -11,6 +11,10 @@ import (
 	"github.com/paketo-buildpacks/packit/internal"
 )
 
+const (
+	ProcessEnvLayerName = "packit-process-launch"
+)
+
 // BuildContext provides the contextual details that are made available by the
 // buildpack lifecycle during the build phase. This context is populated by the
 // Build function and passed to BuildFunc during execution.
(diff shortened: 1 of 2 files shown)


arjun024:
In Draft stage due to ongoing discussions about the layer-process coupling assumed in the spec

joshzarrabi:
I think that makes sense, not too bad to add an extra layer I think?

arjun024:
Retiring this for now

PR #72, opened 2020-09-21
Adds a generic archive type Archive

ForestEckhardt:
- This type will inspect the input steam that it is given and use that
to determine the type of archive the stream is and then use that
information to invoke to correct decompression logic.
- The Archive type has strategies for all currently supported archive
types and should be expanded if more types are added in the future.


vacation/vacation.go
```diff
@@ -133,8 +149,41 @@ func (ta TarArchive) Decompress(destination string) error {
...
+	// https://github.com/gabriel-vasile/mimetype/blob/c64c025a7c2d8d45ba57d3cebb50a1dbedb3ed7e/internal/matchers/matchers.go#L6
+	header, err := bufferedReader.Peek(3072)
+	if err != nil && err != io.EOF {
+		return err
+	}
+
+	mime := mimetype.Detect(header)
+
+	// This switch case is reponsible for determining what the decompression
+	// startegy should be.
```
sophiewigmore, line 170: small typo in strategy

vacation/vacation.go
```diff
@@ -16,9 +17,17 @@ import (
...
+// An Archive decompresses tar, gzip and xz compressed tar, and zip files from
+// an input stream.
+type Archive struct {
+	reader     io.Reader
+	components int
+}
+
 // A TarArchive decompresses tar files from an input stream.
 type TarArchive struct {
 	reader     io.Reader
```
sophiewigmore, line 33: Do we still need TarArchive / NewTarArchive if Archive covers decompressing tar files?

ryanmoran (approved):

PR #75, opened 2020-10-12
Change dash to underscore

ForestEckhardt:
- Changes dash to underscore to reflect the already existing pattern set
by include_files and pre_package

### DIFF
--- a/cargo/config.go
+++ b/cargo/config.go
@@ -29,11 +29,11 @@ type ConfigBuildpack struct {
 }
 
 type ConfigMetadata struct {
-	IncludeFiles    []string                   `toml:"include_files"         json:"include_files,omitempty"`
-	PrePackage      string                     `toml:"pre_package"           json:"pre_package,omitempty"`
-	DefaultVersions map[string]string          `toml:"default-versions"      json:"default-versions,omitempty"`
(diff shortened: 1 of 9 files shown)


ForestEckhardt:
PRs also need to be made against the following repos that set default versions:
- [ ] MRI
- [ ] Bundler
- [ ] PHP-Dist
- [ ] PHP-Composer
- [ ] NGINX
- [ ] Node-engine
- [ ] Yarn
- [ ] Go-dist

ryanmoran:
This differs from what the `libpak` buildpacks do. Do we want to rethink aligning with them for organizational consistency?

ForestEckhardt:
@ryanmoran What do the `libpak` buildpacks do?

ryanmoran:
They use dashes like `pre-package`, `include-files`, and `default-versions`. I get making it consistent, but maybe we should go all the way to aligning more with the rest of the org.

ForestEckhardt:
Ahhhh I thought that the underscores were the standard. I am more than happy to move in the other direction.

ForestEckhardt:
I am going to close this PR and I will open another one different one where I have switched the dashes for underscores.

PR #82, opened 2020-11-02
Don't try to flush stdout when exiting with an error

ryanmoran:
This removes an extra log line that was being printed when the detection for a buildpack failed. Here is an example of what that looked like.

```
===> DETECTING
[detector] ======== Output: paketo-buildpacks/go-build@0.0.36 ========
[detector]
[detector] failed
[detector] ======== Results ========
[detector] fail: paketo-buildpacks/go-build@0.0.36
[detector] ERROR: No buildpack groups passed detection.
[detector] ERROR: Please check that you are running against the correct path.
[detector] ERROR: failed to detect: no buildpacks participating
ERROR: failed to build: executing lifecycle. This may be the result of using an untrusted builder: failed with status code: 100
```

### DIFF
--- a/internal/exit_handler.go
+++ b/internal/exit_handler.go
@@ -50,7 +50,6 @@ func NewExitHandler(options ...Option) ExitHandler {
 }
 
 func (h ExitHandler) Error(err error) {
-	fmt.Fprintln(h.stdout, "")
 	fmt.Fprintln(h.stderr, err)
 
 	var code int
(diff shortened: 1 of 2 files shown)


sophiewigmore:
I tried to reproduce this with the old version of packit to see the extra log line but I didn't see it with any of the examples I ran. What did you do to see it?

sophiewigmore:
Tested this change locally and saw the log line removed.

sophiewigmore (approved):

PR #86, opened 2020-11-13
Adds draft package

ForestEckhardt:
- This package implements a type that is meant to be a generic Buildpack
Plan entry resolves as well as a LayerType merger


judge/plan_entry_handler.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+
+	emitter, ok := p.logger.(Emitter)
+	if ok {
+		emitter.Candidates(filteredEntries)
+	} else {
+		p.candidates(filteredEntries)
+	}
+
+	return filteredEntries[0], true
```
ryanmoran, line 59: I wonder if you could simplify the API of this package by returning just a filtered and sorted list. It reduces some of the cognitive overhead of understanding the meaning attached to the `bool` return value, and negates the need to even deal with a `Logger` or `Emitter` type as those lines can be printed by the caller after they receive a filtered and sorted set of entries.

ForestEckhardt: I think that there are a couple of things that make me hesitant to do this, but maybe I am making too big a deal out of it. Firstly, I like the idea that if you are using this optional function we enforce logging in a format that we feel is appropriate. Secondly, picking an entry feels like it is better for the user in the long run than giving them a list back. If we just hand back a sorted filtered list I feel that there will be a lot of the same code following it which will be to the effect of:
```go
entries := planner.Resolve(name, context.Plan.Entries, priorities)
// bloat
logger.Candidates(entries)
entry := entries[0]
```
I do agree with the idea that the `bool` is not a great interface. Would it maybe make sense to have a typed error must like `os.ErrNotExists` so that it could still be checked?

ForestEckhardt: @ryanmoran if you want to check out my changes and see if that flows better to you.

draft/planner_test.go
```diff
@@ -0,0 +1,271 @@
...
+							"version":        "some-version",
+							"version-source": "highest",
+						},
+					},
+				}))
+			})
+		})
+
+		context("there are no enties matching the given name", func() {
+			it("returns the first entry in the filtered map", func() {
```
sophiewigmore, line 162: I think you just missed this `it`, I think it should be something like:
```suggestion
			it("returns no entries", func() {
```

ryanmoran:
The `dotnet-core-runtime` buildpack is now using a regular expression to evaluate the `version-source` matching a priority declaration: https://github.com/paketo-buildpacks/dotnet-core-runtime/blob/11b7d02648c9749d91356f3c7fa7b1419eca7242/plan_entry_resolver.go#L60-L62

Can we add support for that?

ForestEckhardt:
@ryanmoran There are some issues with doing a regular expression match. I would be interested in hearing what you would propose for the implementation.

ryanmoran:
I've taken a stab at a simple implementation that supports regular expressions: https://github.com/paketo-buildpacks/packit/pull/86/commits/92d1894e186c7592d2c134e258212f9c3f6b383b.

The main design choice was to assume that all strings passed as the priorities are regular expressions.

ryanmoran:
@sophiewigmore maybe [this implementation](https://github.com/paketo-buildpacks/packit/pull/86/commits/668d12c6e932369a7f014c885a64a17272f3c8b1) would be better? In this case, we are allowing strings for exact matching and regexp for fuzzier matching cases.

sophiewigmore:
@ryanmoran Oh wow, yeah. I definitely prefer that to requiring regexp only.

sophiewigmore (approved):

ForestEckhardt:
@ryanmoran @sophiewigmore LGTM

ryanmoran (approved):

PR #89, opened 2020-11-19
Adds update-buildpack command to jam

ryanmoran:
## Summary
The addition of the `jam update-buildpack` command allows buildpack authors to update the `buildpack.toml` and `package.toml` of their language-family buildpacks. The command will query the registry to find new versions of the implementation buildpacks that have been released, updating those files in place.

## Use cases
The command can be run manually, but is likely best used by automation like GitHub Actions to keep buildpacks in a language-family up-to-date.

## Checklist


sophiewigmore:
Code/tests looks good to mean I tested out the functionality and it works great. Can you just reiterate what the need for adding the `docker://` prefix onto the package.toml URIs?

ryanmoran:
@sophiewigmore after discussing the format with the CNB core team, it was made clear that the format we currently use is deprecated and will change to something that looks like this instead:
```toml
[[dependencies]]
  uri = "docker://gcr.io/paketo-buildpacks/bundle-install"
```
I've asked that the team provide better documentation of the file and its format here: https://github.com/buildpacks/docs/issues/261.

As far as code reuse, by plan was to also factor our the builder update code into an `update-builder` command, but I wanted to land this PR first.

sophiewigmore (approved):

PR #90, opened 2020-11-19
fs/sum: implement SumMultiple()

fg-j:
- It allows you to get a digest from multiple inputs.

Signed-off-by: Arjun Sreedharan <asreedharan@vmware.com>

Thanks for contributing. To speed up the process of reviewing your pull request
please provide us with:

* A short explanation of the proposed change:

* An explanation of the use cases your change enables:

Please confirm the following:

### DIFF
--- a/fs/checksum_calculator.go
+++ b/fs/checksum_calculator.go
@@ -1,12 +1,10 @@
 package fs
 
 import (
-	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


ryanmoran:
Why did we go with a new method on the `ChecksumCalculator` instead of modifying the existing `Sum` method to accept variadic arguments? It seemed relatively straight forward to modify the existing `Sum` method to support multiple paths. Here is a hacked up implementation:

```go
func (c ChecksumCalculator) Sum(paths ...string) (string, error) {
	// Collect the paths for all files covered by the set of arguments.
	var files []string
	for _, path := range paths {
		err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
			if err != nil {
				return err
			}

			if info.Mode().IsRegular() {
				files = append(files, path)
			}

			return nil
		})
		if err != nil {
			return "", fmt.Errorf("failed to calculate checksum: %w", err)
		}
	}

	// Sort the list of files to make sure we have a stable list as we shouldn't
	// trust the ordering of filepath.Walk.
	sort.Strings(files)

	//Gather all checksums check for checksum calculation errors
	var sums [][]byte
	for _, f := range getParallelChecksums(files) {
		if f.err != nil {
			return "", fmt.Errorf("failed to calculate checksum: %w", f.err)
		}

		sums = append(sums, f.checksum)
	}

	// NOTE: Special case where we are checksuming a single file. In this case,
	// we return the checksum of the file instead of the checksum of the
	// checksums of the file collection.
	if len(sums) == 1 {
		return hex.EncodeToString(sums[0]), nil
	}

	// Calculate a checksum of the checksum of all files.
	hash := sha256.New()
	for _, sum := range sums {
		_, err := hash.Write(sum)
		if err != nil {
			return "", fmt.Errorf("failed to calculate checksum: %w", err)
		}
	}

	return hex.EncodeToString(hash.Sum(nil)), nil
}
```

ryanmoran (approved):

PR #96, opened 2020-12-08
Ensures layer launch/build/cache flags get cleared on Reset

ryanmoran:
## Summary
The `Layer.Reset` method clears everything except the `Build`/`Launch`/`Cache` flags on the layer. This results in some strange looking code in buildpacks where we can set these flags, then call `Reset` and still have the flags apply later.

## Use Case
This **breaking change** in the API will ensure that these fields are also cleared. That should make code that implements this logic incorrectly break, and hopefully result in buildpacks that have code that is easier to understand.

Please confirm the following:

### DIFF
--- a/layer.go
+++ b/layer.go
@@ -80,21 +80,26 @@ type Layer struct {
 // Reset clears the state of a layer such that the layer can be replaced with
 // new content and metadata. It clears all environment variables, and removes
 // the content of the layer directory on disk.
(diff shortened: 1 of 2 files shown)


ryanmoran:
This is a breaking change. Maybe we can batch this with a few other breaking changes that remove deprecated APIs.

sophiewigmore (approved):

PR #98, opened 2020-12-09
Removes LayerType object and variadic option for Get()

ForestEckhardt:
- This variadic functionality no longer makes sense for Get() as Reset()
now returns a new layer object that does not take into account previous
build layer settings at all. Because the LayerType was only used in this
variadic call the type was also removed.

Please confirm the following:


sophiewigmore:
What if a user doesn't use `Reset`? I feel like it could still be useful to be able to set the `LayerType` when doing `Get`

ForestEckhardt:
@sophiewigmore I feel like `Get` should return the layer as it was restored by the lifecycle. If a user wants to modify the layer further they can do so after retrieving the layer information.

I think if we were to expand the idea of being able to modify things in the `Get` function, like the environment or the metadata. That kinda of functionality doesn't really make sense so why should I be able to modify the layer type during `Get` then.

sophiewigmore:
> @sophiewigmore I feel like `Get` should return the layer as it was restored by the lifecycle. If a user wants to modify the layer further they can do so after retrieving the layer information.
> 
Hmm. I think that I would agree with that if the `Get` method didn't also allow us to create a layer if it doesn't exist. I'm kind of torn. I know you and @ryanmoran had discussed this, was this the consensus you came to in that discussion?

ForestEckhardt:
That it should be removed. It seems strange to keep that artifact around for one scenario of `Get` and the LayerTypes at the end of the day are just for convenience, a user can still set those as properties on the layer.

sophiewigmore (approved):

PR #99, opened 2020-12-10
Removes all APIs that had been previous marked as deprecated

ryanmoran:
## Summary
There have been a number of APIs in the `packit` codebase that have been marked as deprecated for a while now. This removes those deprecated APIs. This is a **breaking change** and will need to include documentation in the release notes about that fact.

Please confirm the following:


cargo/deprecation_warning.go
```diff
@@ -1,33 +0,0 @@
...
-	"github.com/BurntSushi/toml"
-)
-
-type DeprecationWarning struct{}
-
-func NewDeprecationWarning() DeprecationWarning {
-	return DeprecationWarning{}
-}
-
-func (d DeprecationWarning) WarnDeprecatedFields(path string) error {
```
sophiewigmore, line 15: Why do we need to get rid of deprecation warning code? Won't we need it in the future?

ryanmoran: This code was to warn that the `include_files` and `pre_package` fields are deprecated. Now that we've moved all the existing buildpacks over, I don't think we still need this. Do you see any concerns with that?

sophiewigmore: I realize now that this code is hardcoded to look for those fields, so we don't need that anymore. However, I do think we will probably need to reintroduce code like this in the future for deprecations. It might make more sense to re-write this code at that time to be more generic. 

TLDR is I think I'm good getting rid of this now that I realize it's specific to certain deprecated fields.

sophiewigmore (approved):

PR #112, opened 2021-02-02
Adds command for updating builder.toml

ForestEckhardt:
## Summary

A `jam` command that will update a user `builder.toml` to use the latest versions of the buildpacks, lifecycle, and build image.

## Use Cases

Maintaining a `builder.toml`.

Partially solves the following [issue](https://github.com/paketo-buildpacks/github-config/issues/187).

## Checklist

### DIFF
--- a/cargo/jam/internal/image.go
+++ b/cargo/jam/internal/image.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/Masterminds/semver"
 	"github.com/docker/distribution/reference"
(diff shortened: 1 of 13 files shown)


cargo/jam/internal/image.go
```diff
@@ -56,3 +57,59 @@ func FindLatestImage(uri string) (Image, error) {
 		Version: versions[len(versions)-1].String(),
 	}, nil
 }
+
+func FindLatestBuildImage(runURI, buildURI string) (Image, error) {
```
sophiewigmore, line 61: Making sure I understand this - this function is to get the build image that shares the same tag as the run image?

ForestEckhardt: Yes!

sophiewigmore:
Code looks good and mimics what we do in the builder update GHA really well. Tested this out locally and works as expected

ForestEckhardt:
@ryanmoran When you have a chance to take a look!

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Masterminds/semver v1.5.0
+	github.com/Masterminds/semver/v3 v3.1.1
```
ryanmoran, line 8: Do we need 2 different versions of this library?

ForestEckhardt: The v3 version branch allows you to use StrictNewVersion() this means that we won't match against some of the non semver tags in on the lifecycle image (ex. 9832537 in the original branch parses as 9832537.0.0 using the StrictNewVersion() function errors out on this tag). 

I could switch over to using a Regex but this felt more straight-forward

ryanmoran: I guess the question then is why do we need to use the older version elsewhere? I guess I'm just trying to see why we can't just use v3 everywhere.

ForestEckhardt: That's fair. I would just have to go around and convert everything and see if I can just swap everything over.

ForestEckhardt: @ryanmoran Alright, the latest commit should have upgraded everything to the v3 branch.

ryanmoran (approved):

PR #119, opened 2021-02-23
Dependency mappings

### LINKED ISSUE #106: Support dependency mappings
## Context
[RFC-0010](https://github.com/paketo-buildpacks/rfcs/blob/main/accepted/0010-dependency-mappings.md) has been accepted and so we should enable `packit` to support this feature for our buildpacks.

## Proposal
We could extend the functionality of the `postal` package to support looking up dependency mappings if there are `dependency-mapping` bindings available from the platform.

fg-j:
## Summary

Adds the ability to look for dependency mappings in bindings. Notably, doesn't change the interface of the postal Service to avoid disrupting existing buildpacks. The interface will likely change once we fully implement Service Bindings support.

Resolves #106 

## Checklist

### DIFF
--- a/postal/dependency_mappings.go
+++ b/postal/dependency_mappings.go
@@ -8,49 +8,44 @@ import (
 	"strings"
 )
 
-type DependencyMapping struct {
-	SHA256 string
-	URI    string
-}
+type DependencyMappingResolver struct{}
 
-func NewDependencyMapping(sha256, uri string) DependencyMapping {
-	return DependencyMapping{
-		SHA256: sha256,
-		URI:    uri,
-	}
+func NewDependencyMappingResolver() DependencyMappingResolver {
+	return DependencyMappingResolver{}
 }
 
 // Given a target dependency, look up if there is a matching dependency mapping at the given binding path
-func (d Dependency) FindDependencyMapping(bindingPath string) (DependencyMapping, error) {
+func (d DependencyMappingResolver) FindDependencyMapping(sha256, bindingPath string) (string, error) {
+	fmt.Println(bindingPath)
 	allBindings, err := filepath.Glob(filepath.Join(bindingPath, "*"))
 	if err != nil {
-		return NewDependencyMapping("", ""), err
+		return "", err
 	}
 
 	for _, binding := range allBindings {
 		// check if any of the bindings are of type "dependency-mapping"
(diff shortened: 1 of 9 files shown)


postal/dependency_mappings.go
```diff
@@ -0,0 +1,53 @@
...
+		return "", err
+	}
+
+	for _, binding := range allBindings {
+		bindType, err := ioutil.ReadFile(filepath.Join(binding, "type"))
+		if err != nil {
+			return "", fmt.Errorf("couldn't read binding type: %w", err)
+		}
+
+		if strings.Contains(string(bindType), "dependency-mapping") {
```
ryanmoran, line 37:
```suggestion
		if strings.TrimSpace(string(bindType)) == "dependency-mapping" {
```

This should probably be a bit more restrictive so that we don't get a false match with a type like `dependency-mapping-something`.

postal/dependency_mappings.go
```diff
@@ -0,0 +1,53 @@
...
+					return "", err
+				}
+				continue
+			}
+
+			uri, err := ioutil.ReadFile(filepath.Join(binding, sha256))
+			if err != nil {
+				return "", err
+			}
+			return strings.TrimSuffix(string(uri), "\n"), nil
```
ryanmoran, line 49:
```suggestion
			return strings.TrimSpace(string(uri)), nil
```
This should remove any preceding or trailing whitespace.

ryanmoran (changes requested):
As this is implemented, `DependencyMappingResolver` has become part of the public API for `postal`. Do we have a usecase that justifies exposing it as part of the public API, or should we make it private? If it can be justified as a useful part of the public API, then I'd like to see documentation included in the code like we have for the other public types and functions: https://github.com/paketo-buildpacks/packit/blob/e6b39a5f6e87ff08474320a9fbde550b08c25566/postal/service.go#L36-L43

ryanmoran (approved):

ryanmoran:
@fg-j @ForestEckhardt Can you rebase this off `main`? I am getting conflicts that I cannot resolve by just clicking a button.

ryanmoran:
I don't know what happened in the original rebase, but this ended up weirdly including `draft` and `jam update-builder` commits that clearly didn't belong to this PR. I've dropped those commits so its a bit more clear what this PR includes and the history looks more reasonable.

ryanmoran (approved):

PR #123, opened 2021-03-02
Add Optional to BuilderConfigOrderGroup struct

### LINKED ISSUE #122: Jam update-builder cli is removing optional flag from group order entries
## What happened?
When running the jam CLI optional flags are removed from builder group order entries. Additionally empty version tags are introduced when the version was left to "inferred".

Run `jam update-builder --builder-file builder.toml` to update a builder.toml file with the latest versions.

Jam update should follow the https://buildpacks.io/docs/reference/config/builder-config/
 `version` is optional and is included in case of non-empty
 `optional` is optional and is included in case it was set to true. This data currently disappears.

**BEFORE**

```toml
description = "Ubuntu bionic base image with buildpacks for Java, .NET Core, NodeJS, Go, Ruby, NGINX and Procfile"

[[buildpacks]]
  uri = "docker://gcr.io/paketo-buildpacks/ca-certificates:2.1.0"
  version = "2.1.0"

[[buildpacks]]
  uri = "docker://gcr.io/paketo-buildpacks/spring-boot:4.0.0"
  version = "4.0.0"

[lifecycle]
  version = "0.10.2"

[[order]]

  [[order.group]]
    id = "paketo-buildpacks/ca-certificates"
    
  [[order.group]]
    id = "paketo-buildpacks/spring-boot"
    version = "4.0.0"
    optional = true

[stack]
  build-image = "docker.io/paketobuildpacks/build:1.0.22-base-cnb"
  id = "io.buildpacks.stacks.bionic"
  run-image = "index.docker.io/paketobuildpacks/run:base-cnb"
  run-image-mirrors = ["gcr.io/paketo-buildpacks/run:base-cnb"]
```
**AFTER**

```toml
description = "Ubuntu bionic base image with buildpacks for Java, .NET Core, NodeJS, Go, Ruby, NGINX and Procfile"

[[buildpacks]]
  uri = "docker://gcr.io/paketo-buildpacks/ca-certificates:2.1.0"
  version = "2.1.0"

[[buildpacks]]
  uri = "docker://gcr.io/paketo-buildpacks/spring-boot:4.0.0"
  version = "4.0.0"

[lifecycle]
  version = "0.10.2"

[[order]]

  [[order.group]]
    id = "paketo-buildpacks/ca-certificates"
    version = ""
    
  [[order.group]]
    id = "paketo-buildpacks/spring-boot"
    version = "4.0.0"

[stack]
  build-image = "docker.io/paketobuildpacks/build:1.0.22-base-cnb"
  id = "io.buildpacks.stacks.bionic"
  run-image = "index.docker.io/paketobuildpacks/run:base-cnb"
  run-image-mirrors = ["gcr.io/paketo-buildpacks/run:base-cnb"]
```

## Build Configuration
jam v0.7.0 on Linux x64

sophiewigmore:
## Summary

Resolves #122.
Adds the `Optional` field to the `BuilderConfigOrderGroup` struct so that this field isn't lost when the builder is updated.

Omits the `version` field of an order group entry if it's empty, since this is `optional, and inferred`.

## Use Cases

When an order group item is optional, we should persist this information in the new `builder.toml`. If an order group entry has no `version`, then we should leave this as empty rather than overwrite the version.

## Checklist


cargo/jam/commands/update_builder.go
```diff
@@ -49,7 +49,9 @@ func (ub UpdateBuilder) Execute(args []string) error {
 		for j, order := range builder.Order {
 			for k, group := range order.Group {
 				if group.ID == image.Path {
-					builder.Order[j].Group[k].Version = image.Version
+					if builder.Order[j].Group[k].Version != "" {
+						builder.Order[j].Group[k].Version = image.Version
+					}
```
ryanmoran, line 54: I don't understand this change. Does this mean that if I had a version declared in my builder, which we do in almost every case, we wouldn't update it?

sophiewigmore: The version field in the order grouping is optional according to [these docs](https://buildpacks.io/docs/reference/config/builder-config/#order-_list-required_). This change makes it so that if you do not have a version specified, when you run `update-builder`, a version does not get added. The version only gets updated now when the order group entry already had a version.

The original implementation *always* adds a version field with the most current buildpack version, even when the original builder.toml didn't have a version. This isn't a huge issue, but since the field is optional, our script shouldn't be adding a new field.

ryanmoran (approved):

PR #127, opened 2021-03-03
Add support for process launch env. vars.

### LINKED ISSUE #63: Extend packit.Process to include process-specific environment variables

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

## Summary

Adds support for process specific launch env vars.

Fixes #63 

## Use Cases

To allow setting process specific env vars on launch

I see this was previously worked upon in https://github.com/paketo-buildpacks/packit/pull/70

I think it might be simpler to just expose it in the layer instead of in the process struct. This goes well with the rest of env vars that are currently housed in layers and also if someone wants to create a separate layer just for their process - that should work fine. It also allows buildpacks to supply launch env. vars for processes not defined by them.

## Checklist


sophiewigmore:
Hi @samj1912 thank you for the contribution!
I have a question about this implementation. Maybe I'm missing something here - processes can exist independent of a layer. So if `ProcessLaunchEnv` is only exposed in the `Layer` struct, how will we be able to set up process specific environment variables for processes that aren't associated with a layer?

samj1912:
@sophiewigmore IIUC the spec doesn't allow for you to set process specific environment variables without creating a layer. This PR was supposed to simply match packit functionality with the spec. I know this is not ideal but I think there is very little we can do about it right now. This also goes in line with the other env variables that can be set in the layer struct.

I think we should create an upstream issue to allow users to specify env vars for a process without creating an additional layer, but I think meanwhile we should merge this so that atleast we are providing appropriate bindings for the spec.

ryanmoran (approved):
This looks like a reasonable implementation of the current spec. The spec's implementation is unfortunate, but we should still at least support what is currently specified. Thank you @samj1912 for taking the time to implement this.

PR #134, opened 2021-03-10
Adds platform path to build and detect context

ryanmoran:
## Summary

The location of the platform metadata is given to the build and detect processes when they are invoked. This location includes details about user-provided environment variables and other platform extensions that we might want buildpack authors to take advantage of. This addition includes that filesystem location as a parameter on both `DetectContext` and `BuildContext`.

In the process of including this new context, I found that the platform location was hardcoded in the existing `postal.Service.Install` implementation. I've included a new `postal.Service.Deliver` method that allows for the platform path to be given as an argument. The existing `Install` method will reuse the `Deliver` method with a hardcoded platform path to maintain backwards compatibility. The `Install` method has also been marked as deprecated.

Additionally, as I was including a new type that spans both `Detect` and `Build`, it made sense for that type to live in its own file. Once I did that, I thought maybe all the types that aren't directly used in the immediate `Detect` or `Build` functions ought to have their own files to make them a bit easier to find. So, this also includes a bunch of new files that contain those types that have been moved out of `detect.go` and `build.go` into their own files.

## Checklist

### DIFF
--- a/postal/service.go
+++ b/postal/service.go
@@ -3,6 +3,7 @@ package postal
 import (
 	"fmt"
 	"io"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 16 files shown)


sophiewigmore:
Could you explain how having `Platform` as a parameter on the contexts is a benefit? Is it so that we can modify platform fields during detect and build? Why would we need to do this?

postal/service.go
```diff
@@ -129,14 +130,15 @@ func (s Service) Resolve(path, id, version, stack string) (Dependency, error) {
...
-// result.
-func (s Service) Install(dependency Dependency, cnbPath, layerPath string) error {
-	dependencyMappingURI, err := s.mappingResolver.FindDependencyMapping(dependency.SHA256, "/platform/bindings")
+// in a buildpack when packaged for offline consumption can be retrieved. If
+// there is a dependency mapping for the specified dependency, Deliver will use
+// the given dependency mapping URI to fetch the dependency. The dependency is
+// validated against the checksum value provided on the Dependency and will
+// error if there are inconsistencies in the fetched result.
+func (s Service) Deliver(dependency Dependency, cnbPath, layerPath, platformPath string) error {
+	dependencyMappingURI, err := s.mappingResolver.FindDependencyMapping(dependency.SHA256, filepath.Join(platformPath, "bindings"))
```
sophiewigmore, line 141: I didn't realize that the platform path could be anything other than `/platform`, but I need see in the spec that that's just a default value. Random question - at what point could the platform path be set?

ryanmoran: Arguably a platform that isn't `pack` or `kpack` could set a different filepath as the Platform path when it invokes the lifecycle and then hardcoding this like we have wouldn't work.

ryanmoran:
> Could you explain how having `Platform` as a parameter on the contexts is a benefit? Is it so that we can modify platform fields during detect and build? Why would we need to do this?

The CNB spec includes a few extensions that expose information to buildpacks via the Platform path. Having access to that path in a buildpack would be useful for those buildpacks to integrate with those extensions if they needed to.

sophiewigmore:
> The CNB spec includes a few extensions that expose information to buildpacks via the Platform path. Having access to that path in a buildpack would be useful for those buildpacks to integrate with those extensions if they needed to.

Could you point me to an example in the spec, please?

ryanmoran:
The Bindings Extension spec is a good example of this: https://github.com/buildpacks/spec/blob/main/extensions/bindings.md.

sophiewigmore (dismissed):

sophiewigmore (approved):

PR #137, opened 2021-03-12
Add go setup to release creation job

ryanmoran:
## Summary

I'm not entirely sure how this workflow was working before, but we need more control over the version of Go that we are using. I've added a step to install Go 1.16.

## Checklist

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -24,6 +24,10 @@ jobs:
     runs-on: ubuntu-latest
     needs: unit
     steps:
+    - name: Setup Go
+      uses: actions/setup-go@v1
+      with:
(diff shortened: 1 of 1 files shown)


sophiewigmore:
Our other workflows in github-config only set up Go during the unit/integration tests. What part of the release creation step needs a specific Go version?

ryanmoran:
You can see that we've been unable to cut a release since we stopped using `io/ioutil`: https://github.com/paketo-buildpacks/packit/actions/workflows/create-release.yml. The codebase now requires you to use Go 1.16, but that doesn't seem to be what is installed on the action runners and so the compilation fails.

sophiewigmore (approved):

PR #146, opened 2021-03-30
Add bill of materials entry facility for postal.Service

ryanmoran:
## Summary

With the addition of buildpack spec API 0.5 support in recent versions of `packit`, it makes sense to add some convenience functionality that allows buildpack authors to quickly decorate their buildpacks with bill of materials data if they are using `postal`.

## Use Cases

The `postal.Service` type now includes a `GenerateBillOfMaterials` method that will convert a list of `postal.Dependency` values into a valid list of `packit.BOMEntry` values. These can then be included in the `packit.LaunchMetadata` or the `packit.BuildMetadata` as needed.

## Checklist

### DIFF
--- a/postal/buildpack.go
+++ b/postal/buildpack.go
@@ -16,14 +16,8 @@ type Dependency struct {
 	// ID is the identifier used to specify the dependency.
 	ID string `toml:"id"`
 
-	// Version is the specific version of the dependency.
-	Version string `toml:"version"`
-
 	// Name is the human-readable name of the dependency.
 	Name string `toml:"name"`
-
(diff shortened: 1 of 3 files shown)


sophiewigmore:
@ryanmoran Awesome, this will be good to have. Is the idea that this will replace the plan refinery code (ex. https://github.com/paketo-community/pip/blob/main/plan_refinery.go) we have in a lot places when those buildpacks eventually move to buildpack API 0.5?

ryanmoran:
Yes. This is exactly the plan.

sophiewigmore (approved):

PR #157, opened 2021-04-09
Adds function to the emitter that logs launch processes

ForestEckhardt:
## Checklist


scribe/emitter.go
```diff
@@ -83,3 +85,24 @@ Entries:
...
+		p := fmt.Sprintf("%s: %s", process.Type, process.Command)
+
+		if process.Args != nil {
+			p = fmt.Sprintf("%s %s", p, strings.Join(process.Args, " "))
+		}
+
+		// This conditional should be replaced when API 0.6 is supported at which
+		// point any process can be set as default
+		if process.Type == "web" {
+			p = fmt.Sprintf("%s %s", p, "(default)")
```
sophiewigmore, line 102: Is adding the `(default)` after the `web` process necessary? This would be breaking when we use this function in our buildpacks since I don't think we tend to add that now, so I want to make sure we definitely need it. (Looking at `build.go` in the go-build and dotnet-execute buildpacks.)

ForestEckhardt: I think that it will become necessary in the future and would be helpful in some existing buildpack (go-build writes a process for every binary it produces but makes the first binary in the target the default). I think that this is also a little forward thinking because the buildpacks will soon be able to set a process as default and it would be good to reveal that information in the logs.

ryanmoran: I don't think it would break anything in existing buildpacks, but it also isn't strictly necessary. We could opt to just print the processes and omit with one is the default since for API 0.5 and lower that is always web.

ForestEckhardt: @ryanmoran @sophiewigmore What do I need to do to move this PR forward?

ryanmoran: I'd like to removed the `(default)` indicator at this point. We can bring it back in support buildpack api 0.6.

sophiewigmore (approved):

PR #162, opened 2021-04-20
Updates symlink sorting function

ForestEckhardt:
- The existing function had trouble sorting symlinks to symlinks in the
same directory. The new function should sort symlinks that are dependent
on symlinks in the same directory properly

## Checklist

### DIFF
--- a/vacation/vacation.go
+++ b/vacation/vacation.go
@@ -172,28 +172,38 @@ func (ta TarArchive) Decompress(destination string) error {
 	// created before they are created.
 	//
 	// For example:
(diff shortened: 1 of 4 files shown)


ForestEckhardt:
So I misunderstood the vulnerability write up. A symlink can link to any file that exists on the file system, however that symlink must be contained inside of the destination directory. I have change the code such that it better sorts the symlinks for creation in an order such that symlinks of symlinks are created after their base symlink and symlinks can make a link to any file as long as it is and existing file.

ryanmoran (approved):

PR #167, opened 2021-04-30
More flexible fs.ChecksumCalculator

andymoe:
## Summary
Made the following additions to the API in order to allow for adding additional data to the hash after a set of files or directories have been processes. The new API is similar to the [Hash](https://golang.org/pkg/hash/#Hash) interface however other function names may be more clear.

```
calculator := fs.NewChecksumCalculator()

calculator.WritePaths("/my/file", "/my/folder")

// currently Write mirrors Hash.Write signature but we probably don't need the byte length returned
calculator.Write([]bytes{"some string"})
calculator.Write([]bytes{"another string"})

sum := calculator.SumAsHexString()
```

The old API is still around for backwards compatibility: 

```
calculator := fs.NewChecksumCalculator()
sum := calculator.Sum("/my/file", "/my/folder")
```

## Use Case

```
func ShouldRun() (bool, string, error) {
  npm := pexec.NewExecutable("npm")

  buffer := bytes.NewBuffer(nil)
  npm.Execute(pexec.Execution{
      Args:   []string{"config", "get", "node-version"},
      Dir:    "",
      Stdout: buffer,
      Stderr: buffer,
  }))

  calculator := fs.NewChecksumCalculator()
  err := calculator.WritePaths("node_modules")
  if err != nil {
    return false, "", fmt.Errorf("unable to get sha of node_modules for layer caching: %w", err)
  }
  _, err = calculator.Write(buffer.Bytes())
    if err != nil {
    return false, "", fmt.Errorf("unable to determine node version for layer caching: %w", err)
  }

  sum := calculator.SumAsHexString()
  
  cacheSha, ok := metadata["cache_sha].(string)
    if !ok || sum != cacheSha {
    return true, sum, nil
  }
  
  return false, "", nil
}
```

### With no API changes (still viable option I guess)
```
func ShouldRun() (bool, string, error) {
  npm := pexec.NewExecutable("npm")
  
  buffer := bytes.NewBuffer(nil)
  npm.Execute(pexec.Execution{
      Args:   []string{"config", "get", "node-version"},
      Dir:    "",
      Stdout: buffer,
      Stderr: buffer,
  }))

  calculator := fs.NewChecksumCalculator()
  sum, err := calculator.sum("node_modules")
  if err != nil {
     return false, "", fmt.Errorf("unable to get sha of node_modules for layer caching: %w", err)
  }

  hash := sha256.New()
  _, err = hash.Write(buffer.Bytes())
  if err != nil {
     return false, "", fmt.Errorf("unable to get sha of node_modules for layer caching: %w", err)
  }

  _, err = hash.Write([]byte(sum))
    if err != nil {
    return false, "", fmt.Errorf("unable to get sha of node_modules for layer caching: %w", err)
  }

  sum = hex.EncodeToString(hash.sum(nil))

  cacheSha, ok := metadata["cache_sha].(string)
  if !ok || sum != cacheSha {
    return true, sum, nil
  }
  
  return false, "", nil
}
```

## Potential alternative method names

```
calculator := fs.NewChecksumCalculator()

calculator.AddPaths("/my/file", "/my/folder")
calculator.Add([]bytes{"some string"})
calculator.Add([]bytes{"another string"})

sum := calculator.ResultAsHexString()
```


fs/checksum_calculator.go
```diff
@@ -14,11 +15,16 @@ import (
 // ChecksumCalculator can be used to calculate the SHA256 checksum of a given file or
 // directory. When given a directory, checksum calculation will be performed in
 // parallel.
-type ChecksumCalculator struct{}
+type ChecksumCalculator struct {
+	sumValue []byte
+	Hash     hash.Hash
```
ryanmoran, line 20: You probably don't mean to export this. If you did, then including a method on `ChecksumCalculator` that simply mirrors the `Write` method of the `hash.Hash` doesn't seem necessary.

fs/checksum_calculator.go
```diff
@@ -27,8 +33,23 @@ type calculatedFile struct {
...
 
-// Sum returns a hex-encoded SHA256 checksum value of a file or directory given a path.
-func (c ChecksumCalculator) Sum(paths ...string) (string, error) {
+// SumAsHexString returns the sum of the underlying sha256 hash in hex format
+func (c ChecksumCalculator) SumAsHexString() string {
+	return hex.EncodeToString(c.sumValue)
+}
+
+// Write byte slice onto a shared Hash object that it can be summed later
+func (c ChecksumCalculator) Write(p []byte) (n int, err error) {
```
ryanmoran, line 42: Stylistically, I would prefer we didn't use named/bare returns. You'll see that we don't have them anywhere else in the `packit` codebase. Though this method is really small and easy to reason about, they tend to cause confusion about what is actually being returned in any non-trivial method implementation.

ryanmoran (changes requested):
I'd like to understand the intention of an API like this. How would it be used? What use-cases is it solving for?

fs/checksum_calculator.go
```diff
@@ -27,8 +34,30 @@ type calculatedFile struct {
...
+
+// Write a byte slice onto an internal SHA265 hash object so that it can be summed later.
+// This method can be called multiple times and interleaved with calls to
+// `WriteFiles(paths ...string)`
+func (c ChecksumCalculator) Write(p []byte) (int, error) {
+	n, err := c.hash.Write(p)
+	if err != nil {
+		return n, err
+	}
+	return n, err
```
ryanmoran, line 54:
```suggestion
	return c.hash.Write(p)
```

Unless you intend to do something different in the error case, these are logically equivalent.

andymoe: Thanks! I'm debating if it makes sense to hide n entirely is all.

andymoe:
If or when we decide something like this is OK API wise I still think I should backfill tests for the public methods. Right now the new additions are tested indirectly because `Sum` uses them all.

ryanmoran:
I think it could make sense to expose some of these lower-level APIs to the buildpack author, but I think it may make more sense as a new type that is distinct from the `fs.ChecksumCalculator`. Specifically, I think the proposed changes break an implicit contract which is that the `fs.ChecksumCalculator` is stateless. The new API moves state into the calculator and thus makes it difficult to reuse or use in cases of parallel operation. I'd like to propose a different API that would create a new stateful type that can be reused by the `fs.ChecksumCalculator` and therefore keep this implicit stateless contract in-place.

```go
type Checksum struct {
  hash hash.Hash
}

func (c Checksum) WritePaths(paths ...string) error {
  // recursively hashes a filepath, keeping state in c.hash
}

func (c Checksum) Write(b []byte) (int, error) {
  // writes b into c.hash
}

func (c Checksum) EncodeToString() string {
  // returns a hex string encoding of the hash value
}
```

andymoe:
Sounds good. I’ll take a crack at it.

andymoe:
Shall I close this PR in favor of a separate one or reuse this one?

ForestEckhardt:
@andymoe feels like six of one, half a dozen of the other to me. If you feel that a new PR is necessary I would not be offended.

sophiewigmore:
Drafting for now due to inactivity

fg-j:
Closing this for now since it's been inactive for a few months. Feel free to reopen if you still want to contribute this work @andymoe

andymoe:
😭 I’m sorry.

sophiewigmore:
@andymoe Forest carried forward some of this work in https://github.com/paketo-buildpacks/packit/pull/249! So not to worry

ForestEckhardt:
My goal was to review and pick this up next. I'll try and do that later this week!

PR #168, opened 2021-04-30
Adds EnvironmentVariables function to scribe.Emitter

ForestEckhardt:
- This function prints out a formatted map of the build and launch
environment variables and merges the shared environment variables into
both the build and launch environment for printing
- Refactors emitter unit tests to both use occam ContainLines matcher as
well as restructured to follow a pattern of declaring top level
environment variable and doing necessary configuration in it.Before(s)
instead of in the it() blocks themselves. This is an attempt to make the
conditions that the test is testing clearer as you have a well defined
set a parameters for all tests and then the modification is laid out
just before testing.

## Checklist


go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/gabriel-vasile/mimetype v1.2.0
 	github.com/google/go-containerregistry v0.5.0
 	github.com/onsi/gomega v1.11.0
+	github.com/paketo-buildpacks/occam v0.1.2
```
ryanmoran, line 13: So, I didn't think about it at the time you asked me, but this creates a dependency loop. `packit` depends on `occam` which depends on `packit`, and on and on. It means that we will just have releases that create dependabot PRs and thus unreleased commits requiring new releases, ad infinitum.

ForestEckhardt: Hmmm, ok I will take a look at this then. I might just steal the matcher and put it in `packit` as well

ForestEckhardt: I added the ContainLines matcher to the `packit` matchers. If this does not work for you I can look into doing something else.

ryanmoran (approved):

PR #173, opened 2021-05-10
Attempting to remove symlink eval

### LINKED ISSUE #161: packit 0.10.2 unable to extract dependency
## What happened?
See
php-dist: https://github.com/paketo-buildpacks/php-dist/runs/2354116118?check_suite_focus=true
cpython: https://github.com/paketo-community/cpython/runs/2354101672?check_suite_focus=true

Extracting the dependency leads to error with packit 0.10.2.

## Reproduce error
For php-dist:
```
package main

import (
        "log"

        "github.com/paketo-buildpacks/packit/cargo"
        "github.com/paketo-buildpacks/packit/postal"
)

func main() {
        dep := postal.Dependency{
                URI: "https://buildpacks.cloudfoundry.org/dependencies/php/php_7.4.16_linux_x64_cflinuxfs3_87568cc8.tgz",
        }
        s := postal.NewService(cargo.NewTransport())
        err := s.Deliver(dep, "/tmp", "/tmp", "/tmp")
        if err != nil {
                log.Fatal(err)
        }
}
```

```
failed to evaluate symlink /tmp/lib/libgpg-error.so: lstat /tmp/lib/lib: no such file or directory
```

## Checklist

ForestEckhardt:
After doing some more reading and experimenting it look like the symlink sorting and verification that was added to clear CodeQL was not necessary and I can still clear CodeQL without any of that additional work. This also puts us extremely close to our original relatively flexible behavior. 

Resolves #161 

## Checklist


ForestEckhardt:
@ryanmoran @sophiewigmore I now have it in a failing state that I believe should succeed based on my understanding the vulnerability. If you could advise on what change I made that re-introduced the symlink zip slip I would very much appreciate.

sophiewigmore:
@ForestEckhardt in relation to @arjun024 comment on https://github.com/paketo-buildpacks/packit/issues/161, could you work with him to repro the issue he mentioned and verify that these changes fix that issue?

ForestEckhardt:
@sophiewigmore I think that I will only be able to get a meaningful fix out once I understand what is required to pass the scanning for zip slipping. Until then there is not anything super meaningful that can be done.

ForestEckhardt:
The more I read about the symbolic zip slip check the more it sounds like it wants all file that are written and linked to, to be part of the destination directory. If this turns out to be the case I believe that case is unworkable as there are some archives that make symlinks to files on the operating system (see [php-dist](https://github.com/paketo-buildpacks/php-dist)). I would love to be proved wrong about this but I am cynical. If this is not a workable idea how should we proceed forward?

ryanmoran:
My reading of the zipslip vulnerability description is also that mitigation requires a check to ensure all files, symlinks, and symlink targets a within the destination when unarchiving. This would mean that the current implementation should also be doing a check after evaluating the symlink to ensure that the target is within the destination directory.

For the case of `php-dist`, I believe that the current archive file is incorrectly, and therefore unsafely, packaged. We should be using the [`--dereference` flag](https://ftp.gnu.org/old-gnu/Manuals/tar-1.12/html_node/tar_115.html) to be creating truly portable archives, but it appears that we are not.

ForestEckhardt:
The inclination seems to be to close so I will close for now but leave the branch up for a while

Issue #178, opened 2021-05-12
Add update-dependencies command to jam


sophiewigmore:
In issue #140, @ryanmoran outlined a full plan to manage dependency updates in a more transparent manner using the [dependency server](https://github.com/paketo-buildpacks/dep-server). We are breaking that issue down into workable pieces.

This issue is to implement the `update-dependencies` command in the `jam` package.
As described in in #140, the goal of this command is to take in a `buildpack.toml` file that contains `metadata.dependency-constraints` such as:
```
[[metadata.dependency-constraints]]
  id = "bundler"
  constraint = "1.*"
  patches = 2
```
and output an updated `buildpack.toml` if there exists a new version that complies with the given `dependency-constraints`.

The goal here is to have a jam command that can be used in our automation to keep our `buildpack.toml` files up to date according to our constraints. Check out #140 for additional details on the algorithm.

PR #181, opened 2021-05-25
Add jam update-dependencies that uses the dep-server

sophiewigmore:
## Summary

Resolves https://github.com/paketo-buildpacks/packit/issues/178.

`jam update-dependencies` updates a `buildpack.toml` dependencies with the latest versions on the dep-server (api.deps.paketo.io) given a constraint in the `buildpack.toml` that looks like:
```
[[metadata.dependency-constraints]]
  id = "dependency-id"
  constraint = "1.*"
  patches = 2
```

## Use Cases

This will allows us to consume dependencies from the dependency server rather than a hidden Concourse pipeline.

## Checklist

### DIFF
--- a/cargo/config.go
+++ b/cargo/config.go
@@ -30,11 +30,12 @@ type ConfigBuildpack struct {
 }
 
 type ConfigMetadata struct {
-	IncludeFiles    []string                   `toml:"include-files"    json:"include-files,omitempty"`
-	PrePackage      string                     `toml:"pre-package"      json:"pre-package,omitempty"`
-	DefaultVersions map[string]string          `toml:"default-versions" json:"default-versions,omitempty"`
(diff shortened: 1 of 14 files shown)


cargo/config.go
```diff
@@ -69,6 +79,11 @@ func EncodeConfig(writer io.Writer, config Config) error {
 		return err
 	}
 
+	c, err = convertPatches(config.Metadata.DependencyConstraints, c)
```
arjun024, line 82: The Metadata can be technically empty/nil right?

sophiewigmore: Yes but in `convertPatches`, if there is no metadata we just return `c`

sophiewigmore: Even if `config.Metadata` is empty, we can still check `DependencyConstraints`

ForestEckhardt (dismissed):

ForestEckhardt (approved):

PR #186, opened 2021-06-02
Removes useage of OS specfic path separator

### LINKED ISSUE #185: Packit 0.12 fails on Windows
## What happened?
Using a tar.gz-based dependency when running in a Windows buildpack, `packit` incorrectly assumes the TAR file would use Windows path separators `\` ([example](https://github.com/paketo-buildpacks/packit/blob/5674dc46e9f2facabb06f6f3156e28ffb45e3421/vacation/vacation.go#L99)) instead of always using Unix path separators `/` as defined by the tar [spec](https://www.gnu.org/software/tar/manual/html_node/Standard.html#:~:text=separated%20by%20slashes).

Side note: all TARs use `/`, including Windows OCI image layers, only filesystem paths accessed in Windows runtime environments require `\` separators.

Upgrade an unpublished, internal Windows buildpack from packit 0.8.0 to 0.12.0

No-op upgrade

```
Downloading HWC
illegal file path "./": the file path does not occur within the destination directory
```

The dependency we're downloading is here:
```
curl --silent --location https://github.com/cloudfoundry/hwc/releases/download/20.0.0/hwc.tar.gz | tar tvz
drwxr-xr-x  0 root   root        0 Jun 26  2020 ./
-rwxr-xr-x  0 root   root  7058447 Jun 26  2020 ./hwc_x86.exe
-rwxr-xr-x  0 root   root  8579071 Jun 26  2020 ./hwc.exe
```

I assume this is because `packit` will normally ignore a `./` but since it assumes that it should be `.\` instead, it does not ignore it.

## Build Configuration

* What platform (`pack`, `kpack`, `tekton` buildpacks plugin, etc.) are you
  using? Please include a version.
pack @ `main`

* What buildpacks are you using? Please include versions.
Not publicly available.

* What builder are you using? If custom, can you provide the output from `pack
  inspect-builder <builder>`?

* Can you provide a sample app or relevant configuration (`buildpack.yml`,
  `nginx.conf`, etc.)?
```
...
[[metadata.dependencies]]
  id = "hwc"
  name = "Hosted Web Core"
  version = "20.0.0"
  uri = "https://github.com/cloudfoundry/hwc/releases/download/20.0.0/hwc.tar.gz"
  sha256 = "41d0859a1d553ce75dc65b34598dff4e0dd9ebd10750b52ec21f18394cb36800"
  stacks = ["io.buildpacks.stacks.windows.servercore"]
  source = "https://github.com/cloudfoundry/hwc/archive/20.0.0.tar.gz"
  source_sha256 = "643fd1225881bd6206eec205ba818cf60be00bd3a1029c86b0e5bf74a3a978ab"
...
```
## Checklist

ForestEckhardt:
[TAR uses "/" regardless of the OS](https://www.gnu.org/software/tar/manual/html_node/Standard.html#:~:text=separated%20by%20slashes) therefore usage of "\" causes
incorrect behavior.

Resolves #185 

## Checklist

### DIFF
--- a/vacation/vacation.go
+++ b/vacation/vacation.go
@@ -477,10 +477,11 @@ func (z ZipArchive) Decompress(destination string) error {
 
 // This function checks to see that the given path is within the destination
 // directory
-func checkExtractPath(filePath string, destination string) error {
-	destpath := filepath.Join(destination, filePath)
-	if !strings.HasPrefix(destpath, filepath.Clean(destination)+"/") {
-		return fmt.Errorf("illegal file path %q: the file path does not occur within the destination directory", filePath)
+func checkExtractPath(tarFilePath string, destination string) error {
+	osPath := filepath.FromSlash(tarFilePath)
+	destpath := filepath.Join(destination, osPath)
+	if !strings.HasPrefix(destpath, filepath.Clean(destination)+string(os.PathSeparator)) {
+		return fmt.Errorf("illegal file path %q: the file path does not occur within the destination directory", tarFilePath)
 	}
 	return nil
 }
--- a/vacation/vacation.go
+++ b/vacation/vacation.go
@@ -96,7 +96,7 @@ func (ta TarArchive) Decompress(destination string) error {
 		}
 
 		// Skip if the destination it the destination directory itself i.e. ./
-		if hdr.Name == "."+string(filepath.Separator) {
+		if hdr.Name == "./" {
 			continue
 		}
 
@@ -105,7 +105,7 @@ func (ta TarArchive) Decompress(destination string) error {
 			return err
 		}
 
-		fileNames := strings.Split(hdr.Name, string(filepath.Separator))
+		fileNames := strings.Split(hdr.Name, "/")
 
 		// Checks to see if file should be written when stripping components
 		if len(fileNames) <= ta.components {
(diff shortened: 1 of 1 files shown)


ForestEckhardt:
@micahyoung If you wouldn't mind giving this branch a shot and see if it resolves everything for you that would be mighty helpful.

vacation/vacation.go
```diff
@@ -479,7 +479,7 @@ func (z ZipArchive) Decompress(destination string) error {
 // directory
 func checkExtractPath(filePath string, destination string) error {
 	destpath := filepath.Join(destination, filePath)
-	if !strings.HasPrefix(destpath, filepath.Clean(destination)+string(os.PathSeparator)) {
+	if !strings.HasPrefix(destpath, filepath.Clean(destination)+"/") {
```
micahyoung, line 482: I needed one more change to get the comparison right:
```golang
func checkExtractPath(tarFilePath string, destination string) error {
	osPath := filepath.FromSlash(tarFilePath)
	destpath := filepath.Join(destination, osPath)
	if !strings.HasPrefix(destpath, filepath.Clean(destination)+string(os.PathSeparator)) {
		return fmt.Errorf("illegal file path %q: the file path does not occur within the destination directory", tarFilePath)
	}
	return nil
}
```	

Full commit (with yours as a parent) is here: https://github.com/micahyoung/packit/commit/869a1e0b98bc4a2b0dd23d3bdb4d2af3b9775f5a

ForestEckhardt: @micahyoung looks good to me! it has been added and is ready for testings.

micahyoung (approved):
This PR branch is all green on my buildpack! Thanks Forest!

sophiewigmore (approved):

PR #218, opened 2021-08-18
Adds new proposed hash format to BOM Generator

ForestEckhardt:
[Resolves RFC TBD](https://github.com/paketo-buildpacks/rfcs/pull/105)

### DIFF
--- a/postal/service.go
+++ b/postal/service.go
@@ -189,10 +189,16 @@ func (s Service) GenerateBillOfMaterials(dependencies ...Dependency) []packit.BO
 		entry := packit.BOMEntry{
 			Name: dependency.Name,
 			Metadata: map[string]interface{}{
-				"sha256": dependency.SHA256,
-				"source": map[string]string{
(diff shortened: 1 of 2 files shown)


fg-j (commented):
One thing I notice in the RFC is that it explicitly calls out that the possible values for `algorithm` are a finite set:

> As pointed out in the schema, the checksum algorithm should be a
CycloneDX-supported hashing algorithm. The supported options are:
> ```
>   { 'MD5'| 'SHA-1'| 'SHA-256'| 'SHA-384'| 'SHA-512'| 'SHA3-256'| 'SHA3-384'| 'SHA3-512'| 'BLAKE2b-256'| 'BLAKE2b-384'| 'BLAKE2b-512'| 'BLAKE3'}
> ```

Right now, packit's BOMEntry type doesn't enforce this part of the specification. While it's not currently relevant for `postal.Service` since we have SHA256s, I wonder if we can encode this specification in packit to make it easier to enforce in the future.

Issue #107, opened 2021-01-08
Support retrieving bindings by buildpacks


arjun024:
#### Context
Some buildpacks may accept settings/credentials using [bindings](https://github.com/k8s-service-bindings/spec) at build or runtime. A common example would be the host location and credential for an external service.

#### Issue request
Packit should provide a way for a buildpack to look up and retrieve these bindings.

#### Example Use case.

See the session handler buildpacks in this [proposed PHP RFC](https://github.com/paketo-buildpacks/php/blob/a1ef40f8b92906d6086c008124348fa953d38a1e/rfcs/0001-restructure.md). They expect their settings to be provided through a suitable binding.

#### Related links
https://paketo.io/docs/buildpacks/configuration/#bindings
https://github.com/paketo-buildpacks/rfcs/blob/main/accepted/0010-dependency-mappings.md
https://github.com/paketo-buildpacks/maven/tree/v3.2.1#bindings

PR #226, opened 2021-09-13
Adds basic service bindings support

ryanmoran:
## Summary

Implements #107

## Checklist

### DIFF
--- a/extensions/service_bindings_manager.go
+++ b/extensions/service_bindings_manager.go
@@ -1,0 +1,76 @@
+package extensions
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// ServiceBinding is a representation of a service binding that has
+// been provided into the build container.
+type ServiceBinding struct {
+	// Name is the name of the directory that binding was contained within.
+	Name string
+
+	// Type is the content of the type file contained within the binding
+	// directory.
+	Type string
+
+	// Provider is the content of the provider file contained within the binding
+	// directory.
+	Provider string
+
+	// Path is the filepath of the binding directory.
+	Path string
+
+	// Secrets is a mapping of file name to contents for all other files
+	// contained in the bindings directory.
+	Secrets map[string][]byte
+}
+
+// ListServiceBindings returns all service bindings found in the given root
+// directory. These bindings must conform the the Kubernetes Service Binding
+// specification as documented in https://github.com/k8s-service-bindings/spec.
+func ListServiceBindings(root string) ([]ServiceBinding, error) {
+	directories, err := filepath.Glob(filepath.Join(root, "*"))
+	if err != nil {
+		return nil, err
+	}
+
+	var bindings []ServiceBinding
+	for _, directory := range directories {
+		paths, err := filepath.Glob(filepath.Join(directory, "*"))
+		if err != nil {
+			return nil, err
+		}
+
+		binding := ServiceBinding{
+			Name: filepath.Base(directory),
+			Path: directory,
+		}
+
+		for _, path := range paths {
+			content, err := os.ReadFile(path)
+			if err != nil {
+				return nil, err
+			}
+
+			base := filepath.Base(path)
+			switch base {
+			case "type":
+				binding.Type = string(content)
+			case "provider":
+				binding.Provider = string(content)
+			default:
+				if binding.Secrets == nil {
+					binding.Secrets = make(map[string][]byte)
+				}
+
+				binding.Secrets[base] = content
+			}
+		}
+
+		bindings = append(bindings, binding)
+	}
+
+	return bindings, nil
+}
--- a/postal/internal/dependency_mappings.go
+++ b/postal/internal/dependency_mappings.go
@@ -2,9 +2,8 @@ package internal
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
+
+	"github.com/paketo-buildpacks/packit/extensions"
 )
 
 type DependencyMappingResolver struct{}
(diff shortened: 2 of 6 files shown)


extensions/service_bindings_manager.go
```diff
@@ -0,0 +1,76 @@
...
+	// Secrets is a mapping of file name to contents for all other files
+	// contained in the bindings directory.
+	Secrets map[string][]byte
+}
+
+// ListServiceBindings returns all service bindings found in the given root
+// directory. These bindings must conform the the Kubernetes Service Binding
+// specification as documented in https://github.com/k8s-service-bindings/spec.
+func ListServiceBindings(root string) ([]ServiceBinding, error) {
+	directories, err := filepath.Glob(filepath.Join(root, "*"))
```
menehune23, line 34: `ioutil.ReadDir` might be more clear than a custom glob here.

Also, it might be prudent to filter out non-directories here

extensions/service_bindings_manager.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Secrets is a mapping of file name to contents for all other files
+	// contained in the bindings directory.
+	Secrets map[string][]byte
+}
+
+// ListServiceBindings returns all service bindings found in the given root
+// directory. These bindings must conform the the Kubernetes Service Binding
+// specification as documented in https://github.com/k8s-service-bindings/spec.
+func ListServiceBindings(root string) ([]ServiceBinding, error) {
```
menehune23, line 33: Forgive me if this isn't the pattern we've followed for packit, we should consider placing this function into an interface for mockability. Then we can test consumers (e.g. when we add [service bindings to the nginx buildpack](https://github.com/paketo-buildpacks/packit/issues/107)) without needing to write files to the filesystem.

extensions/service_bindings_manager.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	var bindings []ServiceBinding
+	for _, directory := range directories {
+		paths, err := filepath.Glob(filepath.Join(directory, "*"))
+		if err != nil {
+			return nil, err
+		}
+
+		binding := ServiceBinding{
```
menehune23, line 46: Do we care about case-sensitivity for binding names/keys?

postal/internal/dependency_mappings.go
```diff
@@ -22,31 +21,19 @@ func NewDependencyMappingResolver() DependencyMappingResolver {
...
-			return "", fmt.Errorf("couldn't read binding type: %w", err)
-		}
-
-		if strings.TrimSpace(string(bindType)) == "dependency-mapping" {
-			if _, err := os.Stat(filepath.Join(binding, sha256)); err != nil {
-				if !os.IsNotExist(err) {
-					return "", err
-				}
-				continue
+	for _, binding := range bindings {
```
menehune23, line 29: Love that we're able to simplify all this!

extensions/service_bindings_manager.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Path is the filepath of the binding directory.
+	Path string
+
+	// Secrets is a mapping of file name to contents for all other files
+	// contained in the bindings directory.
+	Secrets map[string][]byte
+}
+
+// ListServiceBindings returns all service bindings found in the given root
```
menehune23, line 30: This could be premature, but could it be useful to provide a mechanism to get a single binding (where you know what binding name you're looking for)? That way you don't have to load all the bindings every time you need only a specific one. Since the bindings will presumably be in place for all the buildpacks, I could see a scenario where one buildpack requires a large number of bindings, while another buildpack in the group only needs one or two.

postal/internal/dependency_mappings.go
```diff
@@ -22,31 +21,19 @@ func NewDependencyMappingResolver() DependencyMappingResolver {
 
 // Given a target dependency, look up if there is a matching dependency mapping at the given binding path
 func (d DependencyMappingResolver) FindDependencyMapping(sha256, bindingPath string) (string, error) {
-	allBindings, err := filepath.Glob(filepath.Join(bindingPath, "*"))
+	bindings, err := extensions.ListServiceBindings(bindingPath)
```
menehune23, line 24: Here's another scenario where a targeted API could be nice. Something like:

```go
bindingextensions.GetServiceBinding(bindingPath, "dependency-mapping")
```

extensions/service_bindings_manager.go
```diff
@@ -0,0 +1,76 @@
+package extensions
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// ServiceBinding is a representation of a service binding that has
+// been provided into the build container.
+type ServiceBinding struct {
```
menehune23, line 10: We should consider using `libcnb` where we can. In this case, I think using their binding type would be good: https://github.com/buildpacks/libcnb/blob/main/platform.go#L52

This could get us on a path to eliminate at least a tiny bit of the overhead if we ever try to merge packit with libcnb and/or libpak

menehune23 (changes requested):
In taking a step back at this PR, I'm not so convinced it's necessary in it's current form. This is fully implemented in libpak, and looks easily consumable from there:
https://github.com/paketo-buildpacks/libpak/blob/main/bindings/resolve.go

A couple options:
- Don't merge this, and use that util instead (and possibly other utils in the future, where appropriate), OR
- Make this PR a simple wrapper for that, in case we want it to still be "inside" of packit

sophiewigmore:
@menehune23 @ryanmoran Re: merging our implementation with libcnb/libpak. Are there any risks associated with doing this, given how the two have already diverged from packit a bit? This is something that we could've done in the past with other packit features, and we've opted not to obviously. I worry that using that here will leave packit in a half state where this package uses it, but the rest would need to be refactored a lot to make it consistent.

extensions/service_bindings_manager.go
sophiewigmore: @menehune23 `ioutil` is technically deprecated, even though it hasn't been removed from Go. Because of that, we tend to stay away from it when we can

extensions/service_bindings_manager.go
fg-j: I'm on board with this, given bindings' directory structures can be [pretty elaborate](https://github.com/buildpacks/spec/blob/main/extensions/bindings.md#example-directory-structure). It would be nice to be able to test buildpacks that consume bindings without having to write these structures to the filesystem.

extensions/service_bindings_manager.go
fg-j: It seems like binding names [will always be lowercase](https://github.com/buildpacks/spec/blob/765f2ea3ac1826d6a7abb7aecf496ed2345c1bcd/extensions/bindings.md#detect-and-build-phases), according to the current extension spec.

extensions/service_bindings_manager.go
fg-j: Is your idea here about improving UX for packit consumers or improving build performance/speed?

fg-j (commented):
Like @sophiewigmore, I think it's probably worth having a conversation about packit / libcnb and their futures. *_Perhaps next Working Group meeting?_*

I could see Paketo benefitting from some of the fundamental structs and interfaces defined in libcnb, like [Environment](https://github.com/buildpacks/libcnb/blob/d893d2a8defa3a61a6eb22d20b6fbe018f50ff9d/environment.go) or [Binding](https://github.com/buildpacks/libcnb/blob/8290e25d26d260462daeda2a075479f35412a43b/platform.go#L52), without necessarily having to adopt things like the "Layer Contribution" in libcnb's build function, which I find confusing.

menehune23:
@sophiewigmore @fg-j I spoke extensively with @ryanmoran about my original comments and now have a better understanding why we shouldn't leverage `libpak` just yet. It sounds like there will be larger effort regarding `libcnb` -- spearheaded by the CNB folks and hopefully with input from us -- on how to make `libcnb` more widely useful. It sounds like the hope is that it's followed by some sort of merging of `libpak` and `packit`, which would then heavily leverage `libcnb`.

Until that happens, @ryanmoran was in favor of keeping `libpak` out of `packit` and hence out of this PR. Also, our implementation of secret loading will vary slightly such that it won't make sense to leverage `libcnb` yet here either.

I'm working on an updated PR over at #228

menehune23:
Superseded by #228

PR #228, opened 2021-09-15
Service binding resolver

### LINKED ISSUE #107: Support retrieving bindings by buildpacks

menehune23:
## Summary
Adds a `servicebindings.Resolver` type which can be used to lazily load and resolve service bindings according to the [k8s spec](https://github.com/k8s-service-bindings/spec#workload-projection).

Implementation notes:

- Because the new spec relies on a `$SERVICE_BINDING_ROOT` env var, this PR maintains backwards compatibility by falling back to `$CNB_BINDINGS`, then `<platform>/bindings` if neither env var is set.

- Bindings can be read from either the new [k8s format](https://github.com/k8s-service-bindings/spec#workload-projection) or the [legacy format](https://github.com/buildpacks/spec/blob/main/extensions/bindings.md).

- Rather than providing top-level functions, this PR uses a dedicated type so that consumers can create interfaces for it and also mock/fake it in unit tests to keep them simple (for instance, this eliminates the need to understand the spec's implementation and write files to the filesystem during unit tests that rely on a binding).

- Entry values are of a new `io.ReadCloser`-compatible type, `Entry` -- rather than simply a `string` or `[]byte` -- that can stream the entry from the file system. It also supports `.ReadString()` and `.ReadBytes()` for ease of use. (Credit to @ryanmoran for the great idea).

## Checklist

Closes #107

### DIFF
--- a/postal/fakes/mapping_resolver.go
+++ b/postal/fakes/mapping_resolver.go
@@ -4,11 +4,11 @@ import "sync"
 
 type MappingResolver struct {
 	FindDependencyMappingCall struct {
-		sync.Mutex
+		mutex     sync.Mutex
 		CallCount int
 		Receives  struct {
 			SHA256      string
-			BindingPath string
+			PlatformDir string
 		}
 		Returns struct {
 			String string
@@ -19,11 +19,11 @@ type MappingResolver struct {
 }
 
 func (f *MappingResolver) FindDependencyMapping(param1 string, param2 string) (string, error) {
-	f.FindDependencyMappingCall.Lock()
-	defer f.FindDependencyMappingCall.Unlock()
+	f.FindDependencyMappingCall.mutex.Lock()
+	defer f.FindDependencyMappingCall.mutex.Unlock()
 	f.FindDependencyMappingCall.CallCount++
 	f.FindDependencyMappingCall.Receives.SHA256 = param1
-	f.FindDependencyMappingCall.Receives.BindingPath = param2
+	f.FindDependencyMappingCall.Receives.PlatformDir = param2
 	if f.FindDependencyMappingCall.Stub != nil {
 		return f.FindDependencyMappingCall.Stub(param1, param2)
 	}
--- a/postal/internal/dependency_mappings.go
+++ b/postal/internal/dependency_mappings.go
@@ -2,51 +2,36 @@ package internal
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
+	"github.com/paketo-buildpacks/packit/servicebindings"
 )
 
-type DependencyMappingResolver struct{}
-
-func NewDependencyMappingResolver() DependencyMappingResolver {
-	return DependencyMappingResolver{}
+//go:generate faux --interface BindingResolver --output fakes/binding_resolver.go
+type BindingResolver interface {
+	Resolve(typ, provider, platformDir string) ([]servicebindings.Binding, error)
 }
 
-// Reference file structure for bindings directory
-// - bindings
-//    - some-binding
-//       - type -> dependency-mapping
-// 			 - some-sha -> some-uri
-//       - other-sha -> other-uri
+type DependencyMappingResolver struct {
+	bindingResolver BindingResolver
+}
 
-// Given a target dependency, look up if there is a matching dependency mapping at the given binding path
-func (d DependencyMappingResolver) FindDependencyMapping(sha256, bindingPath string) (string, error) {
-	allBindings, err := filepath.Glob(filepath.Join(bindingPath, "*"))
+func NewDependencyMappingResolver(bindingResolver BindingResolver) DependencyMappingResolver {
+	return DependencyMappingResolver{
+		bindingResolver: bindingResolver,
+	}
+}
+
+// FindDependencyMapping looks up if there is a matching dependency mapping
+func (d DependencyMappingResolver) FindDependencyMapping(sha256, platformDir string) (string, error) {
+	bindings, err := d.bindingResolver.Resolve("dependency-mapping", "", platformDir)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to resolve 'dependency-mapping' binding: %w", err)
 	}
 
-	for _, binding := range allBindings {
-		bindType, err := os.ReadFile(filepath.Join(binding, "type"))
-		if err != nil {
-			return "", fmt.Errorf("couldn't read binding type: %w", err)
-		}
-
-		if strings.TrimSpace(string(bindType)) == "dependency-mapping" {
-			if _, err := os.Stat(filepath.Join(binding, sha256)); err != nil {
-				if !os.IsNotExist(err) {
-					return "", err
-				}
-				continue
-			}
-
-			uri, err := os.ReadFile(filepath.Join(binding, sha256))
-			if err != nil {
-				return "", err
-			}
-			return strings.TrimSpace(string(uri)), nil
+	for _, binding := range bindings {
+		if uri, ok := binding.Entries[sha256]; ok {
+			return uri.ReadString()
 		}
 	}
+
 	return "", nil
 }
--- a/postal/internal/dependency_mappings_test.go
+++ b/postal/internal/dependency_mappings_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 
 	"github.com/paketo-buildpacks/packit/postal/internal"
+	"github.com/paketo-buildpacks/packit/postal/internal/fakes"
+	"github.com/paketo-buildpacks/packit/servicebindings"
 	"github.com/sclevine/spec"
 
 	. "github.com/onsi/gomega"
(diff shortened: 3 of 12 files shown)


bindings/resolver.go
```diff
@@ -0,0 +1,133 @@
...
+	// Type is the type of the binding, given by the content of the 'type' file within the binding directory.
+	Type string
+
+	// Provider is the provider of the binding, given by the content of the 'provider' file within the binding
+	// directory.
+	Provider string
+
+	// Secret is the primary content of the binding. Keys are given by each file name within the binding directory
+	// (other than 'type' or 'provider'), and corresponding values are given by the content of each file.
+	Secret map[string][]byte // TODO: place in custom buffer class
```
menehune23, line 30: Note the TODO: the rhs of this map will implement the `io.Reader` interface

postal/internal/dependency_mappings_test.go
```diff
@@ -66,45 +66,7 @@ func testDependencyMappings(t *testing.T, context spec.G, it spec.S) {
...
-		context("when SHA256 contents cannot be opened", func() {
-			it.Before(func() {
-				Expect(os.MkdirAll(filepath.Join(bindingPath, "some-binding"), 0700)).To(Succeed())
-				Expect(os.WriteFile(filepath.Join(bindingPath, "some-binding", "type"), []byte("dependency-mapping"), 0600)).To(Succeed())
-				Expect(os.WriteFile(filepath.Join(bindingPath, "some-binding", "some-sha"), []byte("dependency-mapping-entry.tgz"), 0000)).To(Succeed())
-			})
-			it("errors", func() {
-				_, err := resolver.FindDependencyMapping("some-sha", bindingPath)
-				Expect(err).To(HaveOccurred())
+				Expect(err).To(MatchError(ContainSubstring("failed to list service bindings")))
```
menehune23, line 69: Haven't updated this file yet...

postal/service.go
```diff
@@ -2,6 +2,8 @@ package postal
 
 import (
 	"fmt"
+	"github.com/paketo-buildpacks/packit/bindings"
```
menehune23, line 5: Will fix ordering

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+	delete(metadata, "provider")
+
+	binding.Secret = metadata
+
+	secrets, err := loadSecrets(filepath.Join(binding.Path, "secret"))
+	if err != nil && !os.IsNotExist(err) {
+		return Binding{}, err
+	}
+	if err == nil {
+		for k, v := range secrets {
```
menehune23, line 230: Technically, this means that any keys within `secret` will trump keys within `metadata`. I don't love it, but it's how existing buildpack implementations handle legacy bindings (via [libcnb](https://github.com/buildpacks/libcnb/blob/main/platform.go#L100)), so it's nothing new.

To handle it differently, there are a couple of options:
- Return two different binding types when resolving: `Binding` or `LegacyBinding` (meh)
- Open up `Binding` to separate `metadata` from `secret` into separate fields (results in unused field for newer binding type and could be confusing)

Overall, it's such an edge case that I don't think it's worth those refactors to address, as it's only for the legacy binding format. No ambiguity for the new format.

Thoughts?

postal/internal/dependency_mappings_test.go
```diff
@@ -32,80 +33,48 @@ func testDependencyMappings(t *testing.T, context spec.G, it spec.S) {
...
-			Expect(os.WriteFile(filepath.Join(bindingPath, "some-binding", "some-sha"), []byte("dependency-mapping-entry.tgz"), 0600)).To(Succeed())
-
-			Expect(os.MkdirAll(filepath.Join(bindingPath, "other-binding"), 0700)).To(Succeed())
-			Expect(os.WriteFile(filepath.Join(bindingPath, "other-binding", "type"), []byte("dependency-mapping"), 0600)).To(Succeed())
-			Expect(os.WriteFile(filepath.Join(bindingPath, "other-binding", "other-sha"), []byte("dependency-mapping-entry.tgz"), 0600)).To(Succeed())
-
-			Expect(os.MkdirAll(filepath.Join(bindingPath, "another-binding"), 0700)).To(Succeed())
-			Expect(os.WriteFile(filepath.Join(bindingPath, "another-binding", "type"), []byte("another type"), 0600)).To(Succeed())
-			Expect(os.WriteFile(filepath.Join(bindingPath, "another-binding", "some-sha"), []byte("entry.tgz"), 0600)).To(Succeed())
+			bindingResolver.ResolveCall.Returns.BindingSlice = []bindings.Binding{
```
menehune23, line 36: No need to write to the file system for this unit test anymore! 🎉

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// SecretReader represents the read-only content of a secret value.
```
menehune23, line 14: Technically this isn't specific to secrets/bindings. Should we move this type out to a more general location in `packit` for use in other places? (renaming to something like `ByteReader` or similar)

postal/internal/fakes/binding_resolver.go
```diff
@@ -0,0 +1,35 @@
+package fakes
```
menehune23, line 1: Generated via `faux`

postal/fakes/mapping_resolver.go
```diff
@@ -7,25 +7,23 @@ type MappingResolver struct {
 		sync.Mutex
 		CallCount int
 		Receives  struct {
-			SHA256      string
-			BindingPath string
+			SHA256 string
```
menehune23, line 10: Generated via `faux`

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, file := range files {
+		content, err := os.ReadFile(filepath.Join(secretsPath, file.Name()))
+		if err != nil {
+			return nil, err
+		}
+		secrets[file.Name()] = MakeSecretReader(content)
```
ryanmoran, line 250: I think I've come around to the idea that we should not read the whole contents into memory before the point at which someone calls `String()`, `Bytes()`, or `Read(b []byte)` on the `SecretReader`. Instead, could the implementation lazily read these contents when those methods are invoked? I think we'd also need to add `Close()` to the API contract to make this work well without exhausting file descriptors in some edgecases.

Basically, I am concerned about a case where a secret is a very large file, like a binary or tarball, that we then read the entire contents of into memory.

bindings/resolver.go
sophiewigmore: I'm not sure how often the collision will occur, but I'm open to the second option you've proposed.  I would rather that the case for the k8s-based binding spec was adhered more strictly (no empty fields) than the legacy-based binding  since that's the direction we are moving in in the long run. 

I would also be okay with leaving the code as is in this case, and then revisiting this if we get user feedback about this being an issue.

bindings/resolver.go
menehune23: Sounds good 👍

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+	// Type is the type of the binding, given by the content of the 'type' file within the binding directory.
+	Type string
+
+	// Provider is the provider of the binding, given by the content of the 'provider' file within the binding
+	// directory.
+	Provider string
+
+	// Secret is the primary content of the binding. Keys are given by each file name within the binding directory
+	// (other than 'type' or 'provider'), and corresponding values are given by the content of each file.
+	Secret map[string]SecretReader
```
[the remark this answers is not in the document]
menehune23: Ah, good question. This is mostly a carryover from the terminology used by the [old spec](https://github.com/buildpacks/spec/blob/main/extensions/bindings.md).

Now that I think about it though, I think even "secret/secrets" is too specific. The newer spec calls them "entries". What do you think of renaming this to `Entries`?

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
+package bindings
```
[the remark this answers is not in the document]
menehune23: Yeah, I ran into that myself (and used `binds` for the variable to resolve it). I considered `servicebindings` as a package name, but it felt long. Happy to change to that or take other suggestions.

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// SecretReader represents the read-only content of a secret value.
+type SecretReader struct {
+	data *bytes.Buffer
+}
```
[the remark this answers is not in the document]
menehune23: No particular reason other than it went along with the `Binding` struct and was small. I'll move it out :)

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// SecretReader represents the read-only content of a secret value.
+type SecretReader struct {
```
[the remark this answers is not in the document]
menehune23: I suppose I chose it because it felt most descriptive of its primary purpose. I feel you though and am not super attached to the name.

Given the other thread about ranaming `Binding.Secret` to `Binding.Entries`, perhaps `Entry` would make sense name here?

bindings/resolver.go
```diff
@@ -0,0 +1,134 @@
...
+			resolved = append(resolved, bind)
+		}
+	}
+	return resolved, nil
+}
+
+// ResolveOne returns a single binding matching the given type and optional provider (case-insensitive). To match on
+// type only, provider may be an empty string. Returns an error if the number of matched bindings is not exactly one, or
+// if there are problems loading bindings from the file system.
+func (r *Resolver) ResolveOne(typ string, provider string) (Binding, error) {
```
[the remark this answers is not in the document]
menehune23: @sophiewigmore I debated this too. I kept it because it's implemented similarly in `libpak`, and I also saw at least one buildpack implementation use it (can't remember off-hand). So I felt like it might be good as an option. I was also planning to use it in the dependency mapping logic here in `packit` but it turned out not to need it.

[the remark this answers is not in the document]
menehune23: I could see it being useful when a buildpack expects only one binding for a given type and wouldn't be able to decide what to do if it were ambiguous anyway.

bindings/resolver.go
menehune23: That shouldn't be difficult to change. I'll look into it. I'll have to move away from using `bytes.Buffer` under the hood and may just use the file object

bindings/resolver.go
menehune23: OK, I'm also in favor of leaving as is. I think if this were a big risk, it would have turned up by now with all the Java buildpacks' implementations (which use `libcnb` under the hood).

bindings/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return &Resolver{
+		bindingRoot: root,
+	}
+}
+
+// Resolve returns all bindings matching the given type and optional provider (case-insensitive). To match on type only,
+// provider may be an empty string. Returns an error if there are problems loading bindings from the file system.
+func (r *Resolver) Resolve(typ string, provider string) ([]Binding, error) {
```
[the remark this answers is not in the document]
menehune23: I know that it's the convention for interface naming (e.g. `Stringer.String()`), but I understand that it may not be as desirable for structs. I just don't know why and have no personal reason for avoiding it (other than sometimes name collisions with interfaces). It also seems hard to avoid when trying to come up with descriptive names.

I'd personally prefer to leave as is, if it's just a nit.

But if you feel more strongly, I'm not opposed to renaming. I think that the names above are perhaps ambiguous as to the plurality (mostly on `Get` as it doesn't necessarily imply one), so if I were to change this, I might suggest something like `Get` and `GetOne`, or `Filter` and `FilterOne`, or something like that.

servicebindings/entry.go
```diff
@@ -0,0 +1,31 @@
+package servicebindings
+
+import "bytes"
+
+// Entry represents the read-only content of a binding entry.
+type Entry struct {
```
menehune23, line 6: @ryanmoran I still need to make this adhere to your lazy-read suggestion

bindings/resolver.go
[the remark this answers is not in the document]
menehune23: Done.

bindings/resolver.go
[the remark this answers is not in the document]
menehune23: Renamed to `servicebindings`

bindings/resolver.go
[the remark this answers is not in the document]
menehune23: Done.

servicebindings/entry.go
menehune23: Done. Interested in your thoughts on the newest commit @ryanmoran

bindings/resolver.go
menehune23: Done

postal/internal/dependency_mappings.go
```diff
@@ -1,52 +1,38 @@
...
 
-// Given a target dependency, look up if there is a matching dependency mapping at the given binding path
-func (d DependencyMappingResolver) FindDependencyMapping(sha256, bindingPath string) (string, error) {
-	allBindings, err := filepath.Glob(filepath.Join(bindingPath, "*"))
+// FindDependencyMapping looks up if there is a matching dependency mapping at the given binding path
+func (d DependencyMappingResolver) FindDependencyMapping(sha256 string) (string, error) {
+	bindings, err := d.bindingResolver.Resolve("dependency-mapping", "")
 	if err != nil {
-		return "", err
+		return "", errors.Wrap(err, "getting dependency-mapping binding")
```
ryanmoran, line 28: Rather than pulling in a whole new package, maybe we can use the stdlib built-in error wrapping?

```suggestion
		return "", fmt.Errorf("failed to resolve dependency-mapping binding: %w", err)
```

postal/service.go
```diff
@@ -39,10 +41,12 @@ type Service struct {
 }
 
 // NewService creates an instance of a Servicel given a Transport.
-func NewService(transport Transport) Service {
+func NewService(transport Transport, platformDir string) Service {
```
ryanmoran, line 44: This changes the public API. Does this mean we want to bump the major version of `packit` as its a backwards-incompatible change?

menehune23: Yes, that's correct. Do we have a way of recording this? (Could be a good case for something like https://github.com/semantic-release/semantic-release eventually)

postal/internal/dependency_mappings.go
menehune23: Done here and elsewhere ✔️

postal/service.go
ryanmoran: I guess I bring this up because we *just* cut a `1.0`. I feel like we should be doing a better job of keeping some stability in the API for a while before we break things again. Is there a way we can make this backwards-compatible?

menehune23: Ah, shoot. I see. Let me look and see if there's a way. The `MappingResolver` interface has changed too, so technically `WithDependencyMappingResolver` isn't backwards compatible either...

menehune23: I wrote all this assuming a major version bump would result, but I didn't realize we just cut a `1.0` :(

ryanmoran: The `MappingResolver` is in an internal package and therefore not part of the public API. I wouldn't worry about that change.

menehune23: True that our mapping resolver implementation is internal, but `WithDependencyMappingResolver` and the `MappingResolver` interface are still public. That said, it would only break folks who are using their own implementation of a mapping resolver, as opposed to ours, so this probably isn't a big deal. In fact, given that, I'm not even sure any library consumer would ever use `WithDependencyMappingResolver` in the first place. I'm guessing we added for DI purposes for our own tests.

menehune23: Here are our options for making this backwards-compatible:
- Hardcode the platform dir here so that we don't need to pass it in. That said, this doesn't actually seem feasible as it is passed in as an argument to build/detect.
- Since the platform dir needs to be fed in somewhere, it seems the only other option is to move it to a parameter on `Resolve`/`ResolveOne`. I could update the lazy-loading logic to store the last used platform dir and only re-read bindings from the filesystem if it has changed.

There may be some other more clever options if you have any other suggestions @ryanmoran

postal/service.go
```diff
@@ -26,9 +28,9 @@ type Transport interface {
 
 //go:generate faux --interface MappingResolver --output fakes/mapping_resolver.go
 // MappingResolver serves as the interface that looks up platform binding provided
-// dependency mappings given a  SHA256 and a path to search for bindings
+// dependency mappings given a SHA256
 type MappingResolver interface {
-	FindDependencyMapping(SHA256, bindingPath string) (string, error)
+	FindDependencyMapping(SHA256 string, platformDir string) (string, error)
```
menehune23, line 33: @ryanmoran This has become the platform dir rather than full binding path. Technically it's a change in the API, but as you stated earlier, the only real implementation of this interface is internal.

postal/service.go
```diff
@@ -175,13 +179,6 @@ func (s Service) Deliver(dependency Dependency, cnbPath, layerPath, platformPath
 	return nil
 }
 
-// Install will invoke Deliver with a hardcoded value of /platform for the platform path.
-//
-// Deprecated: Use Deliver instead.
```
menehune23, line 180: @ryanmoran Is this ok to deprecate now?

postal/service_test.go
```diff
@@ -714,258 +716,6 @@ version = "this is super not semver"
 		})
 	})
 
-	context("Install", func() {
```
menehune23, line 717: At a glance, these looked to be covered by the `"Deliver"` context tests. Feel free to correct me if I missed something.

servicebindings/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+// Resolve returns all bindings matching the given type and optional provider (case-insensitive). To match on type only,
+// provider may be an empty string. Returns an error if there are problems loading bindings from the file system.
+//
+// The location of bindings is given by one of the following, in order of precedence:
+//
+//   1. SERVICE_BINDING_ROOT environment variable
+//   2. CNB_BINDINGS environment variable, if above is not set
+//   3. `<platformDir>/bindings`, if both above are not set
+func (r *Resolver) Resolve(typ string, provider string, platformDir string) ([]Binding, error) {
+	if newRoot := bindingRoot(platformDir); r.bindingRoot != newRoot {
```
menehune23, line 54: Now that the platform is part of this method: If the binding root changes, we'll refresh and cache new bindings. Note that this implicitly occurs on a new resolver, due to the root being initialized to `""`.

postal/service.go
menehune23: @ryanmoran Created a new commit with the second option implemented. Let me know your thoughts. Should be easy to revert/adjust since I left it separate.

postal/service.go
ryanmoran: It is deprecated. Do you mean remove?

postal/service_test.go
ryanmoran: Yeah. This is fine. I always figured they just get removed when we removed the method.

servicebindings/resolver.go
```diff
@@ -0,0 +1,241 @@
...
+
+// Resolve returns all bindings matching the given type and optional provider (case-insensitive). To match on type only,
+// provider may be an empty string. Returns an error if there are problems loading bindings from the file system.
+//
+// The location of bindings is given by one of the following, in order of precedence:
+//
+//   1. SERVICE_BINDING_ROOT environment variable
+//   2. CNB_BINDINGS environment variable, if above is not set
+//   3. `<platformDir>/bindings`, if both above are not set
+func (r *Resolver) Resolve(typ string, provider string, platformDir string) ([]Binding, error) {
```
ryanmoran, line 53: Did you know you can reduce some of the verbosity in these params by declaring types for sequential params that have that same type?

```suggestion
func (r *Resolver) Resolve(typ, provider, platformDir string) ([]Binding, error) {
```

ryanmoran:
I'm happy with where this is heading now. @sophiewigmore do you have anything else you'd like to see before we think about getting this merged?

postal/service.go
menehune23: Oh yes, that's what I meant :)

menehune23: Assuming you're good with this given your comments on the removal of associated tests.

servicebindings/resolver.go
menehune23: Yeah, I know about that but have never preferred it (as it makes modification cumbersome IMO). Happy to conform with the patterns of this repo though.

menehune23: Done here and elsewhere

menehune23:
> I'm happy with where this is heading now. @sophiewigmore do you have anything else you'd like to see before we think about getting this merged?

@ryanmoran @sophiewigmore Just let me know before you merge, and I'll clean up the commit history

sophiewigmore:
@ryanmoran @menehune23 hey y'all. sorry I missed the update earlier. I've been following along with these discussions and this sounds good to me!

ForestEckhardt:
@menehune23 I am still a little unsure about the removal of the `postal.Install` function. I know that it is deprecated but the removal of the function entirely feels like it would necessitate a major bump as we are changing a part of the API by removing functionality. If our goal is to not do a major bump then I feel like that would bee to be added back in unless I am missing something or the intention is to do a major bump and I have just missed the boat on the decision.

menehune23:
> @menehune23 I am still a little unsure about the removal of the `postal.Install` function. I know that it is deprecated but the removal of the function entirely feels like it would necessitate a major bump as we are changing a part of the API by removing functionality. If our goal is to not do a major bump then I feel like that would bee to be added back in unless I am missing something or the intention is to do a major bump and I have just missed the boat on the decision.

@ForestEckhardt That's a good point. Perhaps what I'll do is restore the function but leave the tests removed (since the functionality is covered by `Deliver` tests).

ForestEckhardt (approved):

ForestEckhardt (approved):

PR #230, opened 2021-09-16
Deprecate non-specification field

ForestEckhardt:
## Checklist

### DIFF
--- a/cargo/config.go
+++ b/cargo/config.go
@@ -29,6 +29,8 @@ type ConfigBuildpack struct {
 	Homepage string                   `toml:"homepage,omitempty" json:"homepage,omitempty"`
 	Licenses []ConfigBuildpackLicense `toml:"licenses,omitempty" json:"licenses,omitempty"`
 
+	// Deprecated: This field is not part of the official buildpack.toml spec and
+	// will therefore be removed in the next major release
 	SHA256 string `toml:"-" json:"-"`
 }
 


ryanmoran:
It may not appear to be used in `packit` or by our buildpacks, but `jam` uses this field [here](https://github.com/paketo-buildpacks/jam/blob/d3e4a29ebe31a86a80cb9f61d5b01fe47aea785c/internal/buildpack_inspector.go#L98). Maybe we can figure out how not to use this in `jam` and then we can remove it.

ForestEckhardt:
Interesting because as far as I can figure from the spec the buildpack table has no SHA field.

ryanmoran:
Its not used to read out a value from the `buildpack.toml`. Its just used to hold the value between that point in the `jam` code and the formatter.

ForestEckhardt:
@ryanmoran Here is a PR to remove the use of the `Buildpack.SHA256` field from `jam` https://github.com/paketo-buildpacks/jam/pull/17

ryanmoran:
Are you planning to bump the major version of `packit` when you remove this?

ForestEckhardt:
I hadn't really thought about how I would classify the bump. I think that by definition it would have to be but the field was never parsed it was just part of the structure. It feels a little silly to have a major bump over such a small change.

ryanmoran:
Yes. But strictly speaking, its a backwards-incompatible change. Maybe you should leave the field and just mark it as deprecated which is what we have done in the past. Then, when we get to a point where a major version bump feels warranted because of other, more substantial changes, we can remove it.

ForestEckhardt:
I think that is fair. We can still go forward with the Jam change as well.

ryanmoran (approved):

PR #241, opened 2021-11-04
Remove Paketo-specific BOM from top-level package

### LINKED ISSUE #227: Remove Paketo-specific BOM from top-level package
## Summary
I didn't really think about it at the time, but the new BOM entry types in the top-level `packit` package violate one of the main tenets of the library, which is that the top-level package only implements the spec. We've pushed the structure of the Paketo BOM down into this layer and it means that others consuming this library cannot choose a different BOM structure.

Unfortunately, this ship has sailed and we've released a version of `packit` that already contains this code. Ideally, we would deprecate these new types, create equivalents in a sub-package, and then release a new major version.

I think the changes would be to revert the `BOMEntry` type back to something that looks like the following:
```go
type BOMEntry struct {
	// Name represents the name of the entry.
	Name string `toml:"name"`

	// Metadata is the metadata of the entry.  Optional.
	Metadata interface{} `toml:"metadata,omitempty"`
}
```

With this change, we can use our structured BOM or allow any other value to be used as the metadata.

sophiewigmore:
## Summary

Resolves #227 and also **deprecates usage of this BOM format in favour of https://github.com/paketo-buildpacks/packit/tree/main/sbom**
~~Note, I left references to the `paketosbom.SBOMMetadata` in `build_test` to show examples of how we intend to use it.~~

## Checklist

### DIFF
--- a/bom.go
+++ b/bom.go
@@ -1,84 +1,13 @@
 package packit
 
-import (
-	"fmt"
-	"strings"
-	"time"
-)
-
 // BOMEntry contains a bill of materials entry.
 type BOMEntry struct {
 	// Name represents the name of the entry.
 	Name string `toml:"name"`
 
 	// Metadata is the metadata of the entry.  Optional.
-	Metadata BOMMetadata `toml:"metadata,omitempty"`
+	Metadata interface{} `toml:"metadata,omitempty"`
 }
-
-type BOMMetadata struct {
-	Architecture    string      `toml:"arch,omitempty"`
-	CPE             string      `toml:"cpe,omitempty"`
-	DeprecationDate time.Time   `toml:"deprecation-date,omitempty"`
-	Licenses        []string    `toml:"licenses,omitempty"`
-	PURL            string      `toml:"purl,omitempty"`
-	Checksum        BOMChecksum `toml:"checksum,omitempty"`
-	Summary         string      `toml:"summary,omitempty"`
-	URI             string      `toml:"uri,omitempty"`
-	Version         string      `toml:"version,omitempty"`
-	Source          BOMSource   `toml:"source,omitempty"`
-}
-
-type BOMSource struct {
-	Name            string      `toml:"name,omitempty"`
-	Checksum        BOMChecksum `toml:"checksum,omitempty"`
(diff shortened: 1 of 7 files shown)


fg-j:
Could we name the package bom or sbom instead of paketobom? Or did you explicitly choose this name to imply that there might be other BOMs in packit that don't use Paketo's data structure?

sophiewigmore:
@fg-j I'm definitely open to renaming it, I struggled with this. 

My rationale for not naming it `bom` was that I didn't want anyone to see it and be confused on the difference between the Paketo-specific `bom` package and the top-level spec compliant definition `bom.go`. It should be clear that it's an implementation of the spec. `paketobom` is specific to how we've laid out the Paketo BOM. I'd be up to naming it something else besides `paketobom` or `bom`

sophiewigmore:
@fg-j bump, if you have thoughts here!

fg-j:
@sophiewigmore Given what you explained, I think `paketo-` as the prefix makes sense 🙂 Given recent work in upstream CNB to consistently [refer to BOM as *S*BOM](https://github.com/buildpacks/lifecycle/pull/761), maybe `paketosbom` makes sense. Along the same vein, renaming `bom.go` --> `sbom.go` and the various `BOM`-related structs appropriately.

sophiewigmore:
@fg-j done, everything has been renamed to `sbom` including the underlying TOML field.

sophiewigmore:
@ForestEckhardt I would love to see all of the BOM representation/generation PRs in one release if possible. I think this is something all of the @paketo-buildpacks/tooling-maintainers should align on.

ForestEckhardt:
Oh I see. So continue to preserve this structure for buildpacks that are not on `v0.7` I had not considered that. I think that makes sense!

ryanmoran:
I've rolled back the `BOM` -> `SBOM` conversion as it conflicted with #255. This PR now just consists of moving the old BOM format as-is to a new `paketosbom` package.

sophiewigmore:
Please squash on merge

ryanmoran (dismissed):

ryanmoran (approved):

PR #244, opened 2021-11-10
Updates emitter.LaunchProcesses to use new Default process feild

### LINKED ISSUE #243: Update the `emitter.LaunchProcesses` to display the default process
The `emitter.LaunchProcesses` should display which of the processes is the default process so that user know which start command is going to run. I think that we should also make it more like a formatted table. The following is the format I am proposing:

```
Assigning launch processes:
  foo:  bar (default)
  fred: jane 
```

ForestEckhardt:
- Makes the output more like a formatted table

Resolves #243 
## Checklist

### DIFF
--- a/scribe/emitter.go
+++ b/scribe/emitter.go
@@ -89,11 +89,26 @@ Entries:
 func (e Emitter) LaunchProcesses(processes []packit.Process, processEnvs ...map[string]packit.Environment) {
 	e.Process("Assigning launch processes:")
 
+	var (
+		typePadding int
+	)
+
 	for _, process := range processes {
-		p := fmt.Sprintf("%s: %s", process.Type, process.Command)
+		if len(process.Type) > typePadding {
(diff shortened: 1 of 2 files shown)


sophiewigmore (approved):

ryanmoran:
Didn't get a chance to comment on this PR, but do we think the log line we output now is unclear?

Like, is `(default)` part of the command? Or, what if the command is really long?

```
some-type: some-command
web:            some-other-command (default)
```

```
some-type: some-command
web:            some-other-command with lots of arguments that goes on for such a long thing that its hard to notice that this is the default command unless you are really looking for it (default)
```

Would it make it easier to read if the output looked something like this?
```
some-type:     some-command
web (default): some-other-command
```

ForestEckhardt:
@ryanmoran I am more than happy to make that change! I think that make a good amount of sense I just did not know how to add it. It will complicate the table formatting a bit but I think I can work that out

PR #255, opened 2021-11-19
New package to generate SBOMs using Syft

ryanmoran:
## Summary

Uses the `syft` library to generate SBOMs for arbitrary directory paths and from `postal.Dependency` types.

## Use Cases

This library and the included modifications to the top-level `packit` library will enable buildpack authors to generate SBOMs for their layers:

```go
layer, err := context.Layers.Get("my-layer")
if err != nil {
	return err
}

bom, err := sbom.Generate(layer.Path)
if err != nil {
	return err
}

layer.SBOM.Set("cdx.json", bom.Format(sbom.CycloneDXFormat))
layer.SBOM.Set("syft.json", bom.Format(sbom.SyftFormat))
```

Additionally, buildpack authors using the `postal.Dependency` type can generate an SBOM from that type:

```go
layer, err := context.Layers.Get("my-layer")
if err != nil {
	return err
}

bom, err := sbom.GenerateFromDependency(dependency, layer.Path)
if err != nil {
	return err
}

layer.SBOM.Set("cdx.json", bom.Format(sbom.CycloneDXFormat))
layer.SBOM.Set("syft.json", bom.Format(sbom.SyftFormat))
```

## Checklist


fg-j:
Once this is in a working state and before merging, let's see what it feels like to use this SBOM api in a module buildpack (e.g. node-modules-bom) and a dependency installing buildpack (e.g. yarn). We can integrate feedback from the buildpack-author perspective before putting this into the packit mainline.

robdimsdale:
We have explored what this API looks like for module dependencies for the node-module-bom buildpack (https://github.com/paketo-buildpacks/node-module-bom/pull/37).

Our initial feedback is that the API of providing `packit.SBOMEntries` is a bit confusing and error-prone:
- The map-based interface of `SBOMEntries` effectively maps file extension to file contents.
  - Providing the file extension seems somewhat redundant given that packit already knows which SBOM formats the buildpack supports (via the `sbom-formats` key in the buildpack.toml), and packit already knows the location of the files on the filesystem to write to for each given format.
  - Providing the contents of the formatted SBOM as a byte stream seems necessarily unstructured. We erase all meta information (compared with passing around the formatted SBOMs as `struct`).
- Relating to the above, during our first iteration of implementing against this API, we accidentally tried to re-use the `io.Reader` by copying the result of `bom.Format()` and that silently failed - it wrote an empty byte stream to the second file.

We can take a look at restructuring this API to make it more structured and less error prone. We'll share any updates in this thread.

fg-j:
I've opened a PR against this one with some updates that try to address the feedback @robdimsdale posted above. See https://github.com/paketo-buildpacks/packit/pull/256.

ryanmoran:
Since I've merged #256 into this PR, I'm pulling over some unaddressed comments from that PR:

From @sophiewigmore: https://github.com/paketo-buildpacks/packit/pull/256#issuecomment-982718596
> Looking at the Node Module BOM fork, I like how this API reads in a buildpack. For example, I think because we know a CycloneDX BOM file path should always be `cdx.json` per the spec, that if a buildpack author knows they want that type of BOM, only having to specify `entries.AddFormat(sbom.CycloneDXFormat)`, rather than also specifying the file extension path `cdx.json` is simpler.
> 
> One thing I found slightly confusing was the naming of `sbom.NewEntries`. On first glance, the name made me think we were adding content to a singular BOM file. Maybe naming it something like `entryFormats := sbom.NewEntryFormats(bom)` and then calling `entryFormats.Add(sbom.CycloneDXFormat)` would be slightly clearer for me. Even if you don't go for this naming structure, I find the sbom/entries.go set up slightly confusing and could use some kind of documentation/comments

From @robdimsdale: https://github.com/paketo-buildpacks/packit/pull/256#issuecomment-982873017
> Another +1 for encapsulating the file extension within the format (as shown in this PR). Assuming we're keeping the high-level, agnostic representation in the `packit` package (i.e. a mapping of file extension to file contents) then at least this change makes it easier for consumers of this library, as they don't have to read the spec to determine the file extensions.
> 
> I do recognize that this couples buildpack authors to the `json` file type, and the `syft/spdx/cdx` application types, but I think that's reasonable given that the lifecycle looks for this files specifically, and wouldn't handle other files written by a buildpack that consumes this library.
> 
> My other main feedback is that I would like to use type definitions to describe the intent of the `map[string]io.Reader`? I think it would be more obvious if we had something like below:
> 
> ```
> type FileExtension string
> type FileContents io.Reader
> 
> type SBOMEntries map[FileExtension]FileContents
> ```
> 
> and the calling code remains unchanged.
> 
> However, in my ideal world, I would prefer an extra level of encapsulation here as well - something like this.:
> 
> ```
> type SBomEntry struct {
>     FileExtension String
>     FileContents io.Reader
> }
> 
> type SBOMEntries []SBOMEntry
> ```
> 
> where the calling code then becomes something like this:
> 
> ```
> for entry := range layer.SBOMEntries {
>     err = config.fileWriter.Write(filepath.Join(layersPath, fmt.Sprintf("%s.sbom.%s", layer.Name, entry.FileExtension)), entry.FileContents)
>         if err != nil {
>             panic(err)
>         }
>     }
> ```
> 
> However, I recognize this might feel less streamlined than the `map[FileExtension]FileContents` approach - kind of like forcing Object Oriented principles where they aren't welcome - so I'm willing to compromise here.
> 
> If there are strong objections to both of this suggestions, then at the very least I'd like to see good comments explaining the purpose of the `map[string]io.Reader`.

From @ForestEckhardt: https://github.com/paketo-buildpacks/packit/pull/256#issuecomment-983051448
> Upon some more looking and thinking I would like if we used some keyword other that `entry`/`entries` as I feel it might overload the term with is already used for entries from the buildpack plan. Unless I am missing something and that is the upstream terminology for those pieces of information.

ryanmoran:
I've trimmed the API down to something that looks like this:

```go
func Build() packit.BuildFunc {
    layer, err := context.Layers.Get("my-layer")
    if err != nil {
      return packit.BuildResult{}, err
    }

    bom, err := sbom.Generate(layer.Path)
    if err != nil {
      return packit.BuildResult{}, err
    }

    layer.SBOM = bom.EntriesFor(context.BuildpackInfo.SBOMFormats...)

    return packit.BuildResult{
      Layers: []packit.Layer{layer},
    }, nil
  }
}
```

So, the `SBOM` field on the `Layer`, `BuildMetadata`, and `LaunchMetadata` types now accept a type with the following interface:
```go
Format() map[string]io.Reader
```

This comes directly from @fg-j's contributions and should address concerns about the accidental reuse of an `io.Reader`. 

At the same time, I've removed a bunch of duplication in the `sbom` package so that the current API looks like the following:

```go
type SBOM struct {}
func Generate(path string) (SBOM, error)
func GenerateFromDependency(dependency postal.Dependency, path string) (SBOM, error)
func (s SBOM) EntriesFor(formats ...Format) Entries

type Entries map[string]io.Reader
func (e Entries) Format() map[string]io.Reader

type Format string
const (
  CycloneDXFormat Format = "application/vnd.cyclonedx+json"
  SPDXFormat      Format = "application/spdx+json"
  SyftFormat      Format = "application/vnd.syft+json"
)
func (f Format) Extension() string

type Formatter struct {}
func NewFormatter(s SBOM, f Format) *Formatter
func (f *Formatter) Read(b []byte) (int, error)
```

I think this addresses a bit of the previously mentioned concerns while keeping the API streamlined. However, there are still concerns that are left unaddressed at this point. Specifically,
* Naming of the type that holds the file-to-content mapping, `Entries` is possibly ambiguous
* Making the `map[string]io.Reader` clearer as to its purpose

Feel free to try this out and provide further feedback as well.

fg-j:
This interface feels pretty good to me. I took a stab at renaming to address the "Entries" overloading. Now the word "format" just sounds like nonsense in my head, but I think these names are a bit more clear about each component's role.

robdimsdale:
This is much nicer. Great work everyone! I think we should rename the `sbom_entries.go` file to `sbom_formatter.go` to match the rename @fg-j did. We'll also need to update the `go generate ...` line to match the new interface.

I would do these changes myself, but it would require opening another PR as I can't commit directly to this branch.

ryanmoran:
I'm working on documenting the new package. I'll make @robdimsdale's proposed changes at the same time.

sophiewigmore:
The renamings have satisfied my open request!

sbom_formatter.go
```diff
@@ -0,0 +1,26 @@
...
+import "io"
+
+// SBOMFormat represents the mapping of a formatted SBOM content to its file
+// extension on the filesystem.
+type SBOMFormat struct {
+	Extension string
+	Content   io.Reader
+}
+
+// SBOMFormatter defines the interface for
```
fg-j, line 12: Are you missing the back half of a sentence here?

fg-j:
This API looks good to me. I like the addition of `packit.SBOMFormats` for inlining slices of formatted SBOMs. In a buildpack, usage looks like
```go
inline := packit.SBOMFormats{
  {
    Extension: "spdx.json",
    Content:   bytes.NewReader([]byte(`{ "hello" : "world" }`)),
  },
  {
    Extension: "cdx.json",
    Content:   bytes.NewBuffer([]byte(`{ "goodbye" : "moon" }`)),
  },
}
```

sbom_formatter.go
ryanmoran: LOL

robdimsdale:
One other change that I would love to see if we're open to it, is the return of something we saw in earlier iterations - getting formats directly from the `BuildpackInfo.SBOMFormats` rather than manually creating the list:

Specifically we used to be able do this:
```
    layer.SBOM = bom.EntriesFor(context.BuildpackInfo.SBOMFormats...)
```

But that won't work directly anymore, as we're now using a type for the Format, rather than a string. But, we could make this work if we also provided a couple of helper functions in `sbom/formats.go` e.g.

```
func FormatsFromStrings(s []string) ([]Format, error) {
	formats := []Format{}
	for _, possibleFormat := range s {
		format, err := FormatFromString(possibleFormat)
		if err != nil {
			return nil, err
		}
		formats = append(formats, format)
	}
	return formats, nil
}

func FormatFromString(s string) (Format, error) {
	switch s {
	case string(CycloneDXFormat):
		return CycloneDXFormat, nil
	case string(SPDXFormat):
		return SPDXFormat, nil
	case string(SyftFormat):
		return SyftFormat, nil
	default:
		panic(fmt.Sprintf("invalid format: %s", s))
	}
}

```

This kind of feels clunky, but I do like the idea of relying on the values in the `BuildpackInfo.SBOMFormats` rather than hard-coding that list in the buildpack's `BuildFunc`.

Thoughts?

fg-j:
@robdimsdale Rather than add more API surface, what about switching `sbom.InFormats()` to accept strings instead Formats, and handling the conversion inside `InFormats()`? We could use an unexported `stringToFormat()` function. Something like:
```go
func (s SBOM) InFormats(formats ...string) (Formatter, error) {
	var fs []Format
	for _, f := range formats {
		format, err := stringToFormat(f)
		if err != nil {
			return Formatter{}, err
		}
		fs = append(fs, format)
	}
	return Formatter{sbom: s, formats: fs}, nil
}

func stringToFormat(format string) (Format, error) {
	switch format {
	case string(CycloneDXFormat):
		return CycloneDXFormat, nil
	case string(SPDXFormat):
		return SPDXFormat, nil
	case string(SyftFormat):
		return SyftFormat, nil
	default:
		return "", fmt.Errorf("%s is not a supported SBOM format", format)
	}
}
```

robdimsdale:
@fg-j that works too. The main downside I see for this approach is that we provide a type definition (`type Format string`) but then don't actually ask consumers to use it - instead asking them to use the underlying `string`. That doesn't seem great. I think that the fact that `BuildpackInfo.SBOMFormats` doesn't have more type information than just `string` shouldn't bleed into the API here.

I think I'm happy with either your solution, my solution, or ignoring it all together and letting users hard-code the BOM types in the `BuildFunc`. It's not like they will change often.

ryanmoran:
I like relaxing the API for `sbom.SBOM.InFormats`. @fg-j, I think we can simplify your implementation a bit though:

```go
func (s SBOM) InFormats(formats ...string) (Formatter, error) {
	var fs []Format
	for _, f := range formats {
		format := Format(f)
		if format.Extension() == "" {
			return "", fmt.Errorf("%q is not a supported SBOM format", f)
		}
		fs = append(fs, format)
	}
	return Formatter{sbom: s, formats: fs}, nil
}
```

ryanmoran:
@robdimsdale let me know how this new version works for you.

robdimsdale:
This is great. Let's merge it.

As a bonus, I notice that the upstream PR in syft (https://github.com/anchore/syft/pull/635) has been merged and a new release of syft has been created (v0.31.0). So, before we merge this, can we remove the `replace` directive in `go.mod`?

robdimsdale:
@ryanmoran there are some other changes that we'll need to make as a result of the changes in upstream syft between when the CycloneDX PR was created and when it was merged. I'll work on those and create a PR against this PR. I'll also address the `go.mod` issues I describe above.

robdimsdale:
https://github.com/paketo-buildpacks/packit/pull/262 contains the changes to account for the upstream PR being merged.

ForestEckhardt (approved):

go.mod
```diff
@@ -5,16 +5,15 @@ go 1.16
...
 	github.com/pelletier/go-toml v1.9.4
 	github.com/sclevine/spec v1.4.0
 	github.com/ulikunitz/xz v0.5.10
-	golang.org/x/net v0.0.0-20210805182204-aaa1db679c0d // indirect
-	golang.org/x/sys v0.0.0-20210809222454-d867a43fc93e // indirect
-	gopkg.in/check.v1 v1.0.0-20200902074654-038fdea0a05b // indirect
 )
+
+// TODO: remove once a new release is cut with the CycloneDX format (probably v0.32.0)
+replace github.com/anchore/syft => github.com/anchore/syft v0.31.1-0.20211204010623-5374a1dc6ff6
```
fg-j, line 19: are we happy merging this into main?

fg-j: the alternative, i guess, is blocking on a syft release...

sophiewigmore (approved):

go.mod
sophiewigmore: I'm OK with it

sophiewigmore: The other alternative is to add our own JSON conversion code, which just doesn't seem worth it at this point.

fg-j (approved):

ForestEckhardt (approved):

PR #256, opened 2021-11-24
Updated SBOM Interface

fg-j:
## Summary

This update tries to
- maintain a very generic interface in the root package (still relying on an `io.Reader`) to allow buildpack authors flexibility on SBOM implementations)
- reuse as much of the prior implementation as possible
- eliminate buildpack authors' ability to accidentally re-use a reader within buildpack code
- add a bit of debugability to `Entries` by providing a `GetContent()` function that can be used to print SBOMs during buildpack development

There are still lots of `panic()` in there and a few Pending unfinished tests. Feedback is welcome.

## Use Cases

With this interface, a buildpack author would do things like:
```go
	layer, err := context.Layers.Get("my-layer-name")

	layer, err = layer.Reset()

	bom, err := sbom.Generate(context.WorkingDir)

	entries := sbom.NewEntries(bom)
	entries.AddFormat(sbom.CycloneDXFormat)
	entries.AddFormat(sbom.SyftFormat)
	entries.AddFormat(sbom.SPDXFormat)

	layer.SBOM = entries

	b, err := ioutil.ReadAll(entries.GetContent(sbom.CycloneDXFormat))
	fmt.Printf("bom content:\n%s\n", string(b))

	return packit.BuildResult{
		Layers: []packit.Layer{
			layer,
			},
		Build: packit.BuildMetadata{},
			SBOM: entries,
		},
        }, nil
```

For an runnable example buildpack implementation that uses this updated interface, see [this fork of the node modules bom buildpack](https://github.com/robdimsdale/node-module-bom/tree/sbom-syft-fgj).

## Checklist


fg-j:
@paketo-buildpacks/tooling-maintainers please take a look 🙂

ForestEckhardt:
I have no issues with this as it stands. I have a question for the larger group around potential extension. I am not super thrilled with this structurer:
```go
entries := sbom.NewEntries(bom)
entries.AddFormat(sbom.CycloneDXFormat)
entries.AddFormat(sbom.SyftFormat)
entries.AddFormat(sbom.SPDXFormat)
```

I do not think that understandability is gained by having separate calls like this, I would love to see something like:
```go
entries := sbom.NewEntries(bom)
entries.AddFormats(sbom.CycloneDXFormat, sbom.SyftFormat, sbom.SPDXFormat)
```

Where the signature of `AddFormats` is `AddFormats(formats ...Format)`.

If this feels premature I think that is fair just wanted to put it forward as it feels like something I would add to go in the next release I could get it into.

fg-j:
cc @thitch97 here's the alternate/amended proposal

ryanmoran:
I'm not sure I understand this statement:

> eliminate buildpack authors' ability to accidentally re-use a reader within buildpack code

`Reader` types, as a general rule, are not reusable unless they provide some sort of "rewind" capability. For example, https://go.dev/play/p/5xJW7ZS0ScA. This makes some sense though. A reader isn't a stateless bag of bytes. Its a stateful stream of bytes with a cursor position. If one bit of your code moves the cursor to the end of the stream by reading the whole thing, then later bits should not expect to be able to re-read that same stream without it somehow being rewound to the beginning. Some `Reader` types are backed by implementations that provide a `Seek(offset int64, whence int) (ret int64, err error)` method. Using that, you could rewind back to the beginning.

You could maybe make an argument that the `Format` method of the `sbom.SBOM` type should return a `ReadSeeker`. And then `packit` would make sure to rewind the reader back to the beginning before reading it. This would make for a pretty cumbersome API though. There are lots of alternative `Format` implementations that wouldn't reasonably be able to return a type capable of rewinding the `Reader`. `bytes.Buffer` stands out as a very commonly used `Reader` implementation for shuttling bytes between types. Unfortunately, it doesn't support the `Seeker` interface. `packit` users would be forced to wrap these types in something capable of providing that `Seeker` interface. To me this feels like a big penalty to the majority of `packit` users for a use-case that seems like a small minority (reusing the BOM in multiple places).

If users understand that the `Reader` interface is stateful, then reusing the BOM simply becomes ensuring that you recall the `sbom.SBOM.Format` method to get a new instance of a reader in each location you'd like a formatted BOM. For example,

```go
layer, err := context.Layers.Get("my-layer")
if err != nil {
	return err
}

bom, err := sbom.Generate(layer.Path)
if err != nil {
	return err
}

layer.SBOM.Set("cdx.json", bom.Format(sbom.CycloneDXFormat))
layer.SBOM.Set("syft.json", bom.Format(sbom.SyftFormat))

return packit.BuildResult{
	Layers: []packit.Layer{layer},
		SBOM: packit.SBOMEntries{
			"cdx.json":  bom.Format(sbom.CycloneDXFormat),
			"syft.json": bom.Format(sbom.SyftFormat),
		},
	},
}, nil
```

fg-j:
@ryanmoran I appreciate you diving into the technical reasons why a `ReadSeeker` probably doesn't make sense as the `sbom.SBOM.Format()` return type. I wholeheartedly agree.

Hopefully I can shed some light on what it felt like to use the v1 proposed packit implementation to rewrite the node-modules-bom buildpack. That experience directly motivated this PR. @robdimsdale and I did look at `packit` and see that `bom.Format()` would return a `Reader`. And we did understand that a `Reader` isn't reusable. However, along the way, we ended up writing some code that seemed benign (and maybe even semantically correct) that ran right into the non-reusability:
```go
layer.SBOM.Set("cdx.json", bom.Format(sbom.CycloneDXFormat))
layer.SBOM.Set("syft.json", bom.Format(sbom.SyftFormat))

return packit.BuildResult{
	Layers: []packit.Layer{layer},
		SBOM: layer.SBOM
		},
	},
}, nil
```

We `Set` some contents of the layer SBOM, and then we set the Launch SBOM equal to the layer SBOM. In retrospect, it's obvious why this was problematic, but in the moment we reasoned that, semantically, we needed to put the same SBOM contents in both of these places. It took stepping through packit's `Build()` to be reminded about why our buildpack was broken.

In the process of debugging the SBOM, we also wanted to print out the contents of the SBOM at build time. Printing the SBOM struct itself doesn't yield anything useful because its content is in the private `sbom` field. In this case, we were tempted to "really see" what's in the layer SBOM by reading one of the values in the map. This also emptied the reader.

In both of these cases, code that feels intuitive to write produces a broken result. 

In this proposal, I started from an API that would have been most intuitive for me to write and read as a buildpack author and then tweaked `packit` to iteratively approach that state.

@ryanmoran What concerns do you have with the proposal in this PR?

fg-j:
@robdimsdale I think you had other thoughts about the API besides what I mentioned above. Care to expound?

sophiewigmore:
Looking at the Node Module BOM fork, I like how this API reads in a buildpack. 
I think because we know a CycloneDX BOM should always be `cdx.json` per the spec, that if a buildpack author knows they want that type of BOM, only having to specify`entries.AddFormat(sbom.CycloneDXFormat)`, rather than also specifying the file extension path `cdx.json` is simpler. 

One thing I found slightly confusing was the naming of `sbom.NewEntries`. On first glance, the name entries makes me think of adding content to a singular BOM file. Maybe naming it something like `entryFormats := sbom.NewEntryFormats(bom)` and then calling `entryFormats.Add(sbom.CycloneDXFormat)` would be slightly clearer for me.

robdimsdale:
Another +1 for encapsulating the file extension within the format (as shown in this PR). Assuming we're keeping the high-level, agnostic representation in the `packit` package (i.e. a mapping of file extension to file contents) then at least this change makes it easier for consumers of this library, as they don't have to read the spec to determine the file extensions.

I do recognize that this couples buildpack authors to the `json` file type, and the `syft/spdx/cdx` application types, but I think that's reasonable given that the lifecycle looks for this files specifically, and wouldn't handle other files written by a buildpack that consumes this library.

My other main feedback is that I would like to use type definitions to describe the intent of the `map[string]io.Reader`? I think it would be more obvious if we had something like below:

```
type FileExtension string
type FileContents io.Reader

type SBOMEntries map[FileExtension]FileContents
```

and the calling code remains unchanged.

However, in my ideal world, I would prefer an extra level of encapsulation here as well - something like this.:

```
type SBomEntry struct {
    FileExtension String
    FileContents io.Reader
}

type SBOMEntries []SBOMEntry
```

where the calling code then becomes something like this:

```
for entry := range layer.SBOMEntries {
				err = config.fileWriter.Write(filepath.Join(layersPath, fmt.Sprintf("%s.sbom.%s", layer.Name, entry.FileExtension)), entry.FileContents)
				if err != nil {
					panic(err)
				}
			}
```

However, I recognize this might feel less streamlined than the `map[FileExtension]FileContents` approach - kind of like forcing Object Oriented principles where they aren't welcome - so I'm willing to compromise here.

If there are strong objections to both of this suggestions, then at the very least I'd like to see good comments explaining  the purpose of the `map[string]io.Reader`.

ryanmoran:
@fg-j I would like to merge this into the other PR and then make some further changes from that point. Does that sounds ok?

ForestEckhardt:
Upon some more looking and thinking I would like if we used some keyword other that `entry`/`entries` as I feel it might overload the term with is already used for entries from the buildpack plan. Unless I am missing something and that is the upstream terminology for those pieces of information.

robdimsdale:
I have no attachment to the specific naming of `entry`. Naming is hard.

ForestEckhardt:
> Naming is hard.

Oh yeah

fg-j:
@ryanmoran Go for it!

Issue #179, opened 2021-05-13
Simplify buildpack.toml encoding/decoding


sophiewigmore:
The code in [cargo/config.go](https://github.com/paketo-buildpacks/packit/blob/main/cargo/config.go) is responsible for decoding `buildpack.toml` files into Go structs, as well was encoding structs into TOML. This code is integral to the `jam` tool.

There are a couple of concerns around this code, given how integral it is to our tools.
1. The TOML decoding/encoding is done with the [`BurntSushi/toml`](https://github.com/BurntSushi/toml) package, which is in an unmaintained/archived status. It's worrisome to use a package in this state.
2. Due to the nature of [the structs](https://github.com/paketo-buildpacks/packit/blob/main/cargo/config.go#L11-L58) involved and the usage of the `BurntSushi/toml` package we have to convert into JSON as an intermittent step in order to decode/encode correctly. This makes for extremely complicated and hard to maintain code.

While all of this functionality works for our use case, this issue is here to document the technical debt associated with this code.

There is a newer TOML library, [github.com/pelletier/go-toml](https://github.com/pelletier/go-toml) that looks like a promising alternative for our use case. Upon initial investigation, there are a couple outstanding issues around using custom TOML marshalling that prevent us from using it currently. Upon the fix of theses issues this could be an option to resolve this.

fg-j:
@sophiewigmore Is this still blocked on improvements to pelletier/go-toml? If so, can you link to the outstanding issues? If not, feel free to remove the `status/blocked` label.

fg-j:
Related to this issue, I notice that [`buildpack_info.go`](https://github.com/paketo-buildpacks/packit/blob/main/buildpack_info.go) and [`cargo/config.go`](https://github.com/paketo-buildpacks/packit/blob/main/cargo/config.go) both contain structs that represent the contents of a `buildpack.toml`. @paketo-buildpacks/tooling-maintainers , why do we have two separate representations? When the spec for `buildpack.toml` contents changes, we end up having to make changes in two places, as with [recent SBOM work](https://github.com/paketo-buildpacks/packit/pull/248).

sophiewigmore:
@fg-j Interestingly enough, BurntSushi doesn't appear to be unmaintained or archived anymore, so this is less of a worrisome issue. However, the issue of complicated code still stands.

[...]

PR #261, opened 2021-12-03
Use packit.BuildpackInfo inside cargo.Config

sophiewigmore:
## Summary

In https://github.com/paketo-buildpacks/packit/issues/179#issuecomment-983086623, @fg-j asks why we have duplication across `packit.BuildpackInfo` and `cargo.ConfigBuildpack`. 

I also can't see why we need to have both of these structs, so I have removed `cargo.ConfigBuildpack` and modified `packit.BuildpackInfo` to contain the appropriate `json` labels. 

If there is a reason for the duplication, I would like an explanation so we have that historical context in place going forward.

## Checklist

### DIFF
--- a/buildpack_info.go
+++ b/buildpack_info.go
@@ -6,35 +6,39 @@ package packit
 type BuildpackInfo struct {
 	// ID is the identifier specified in the `buildpack.id` field of the
 	// buildpack.toml.
-	ID string `toml:"id"`
+	ID string `toml:"id"                    json:"id,omitempty"`
 
 	// Name is the identifier specified in the `buildpack.name` field of the
 	// buildpack.toml.
-	Name string `toml:"name"`
+	Name string `toml:"name"                  json:"name,omitempty"`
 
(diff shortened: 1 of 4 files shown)


ryanmoran:
I think I'm fine with this in theory, but I don't like that we ended up with changes to the top-level `packit` library (the addition of the JSON tags) that only need to be there to support some internal code that lives in `cargo`. I think this would make a lot more sense after we resolve the code complexity in `cargo` that forces us to convert to JSON and then back to TOML.

sophiewigmore:
@ryanmoran I hear that! I wasn't sure if the JSON fields would really matter much but I think that's a valid concern. Let me draft this PR until I've had a chance to look at go-toml v2

sophiewigmore:
I investigated using the [go-toml v2](https://github.com/pelletier/go-toml/tree/v2) beta release to skip the JSON conversion step, but I couldn't get it into a working state. Closing this PR out until we can figure out a solid way to encode/decode the `cargo.Config` type.

PR #272, opened 2021-12-23
Simplify SBOM test assertions

fg-j:
## Summary

I have noticed over the past weeks that our SBOM tests were quite brittle – frequently requiring intervention when upgrading syft versions – despite there not being important changes to the outputted JSON. I took a shot at stripping down the assertions to minimize the amount we're just testing syft's functionality.

## Checklist

### DIFF
--- a/sbom/sbom_test.go
+++ b/sbom/sbom_test.go
@@ -46,24 +46,7 @@ func testSBOM(t *testing.T, context spec.G, it spec.S) {
 				}
 			}
 
-			type artifact struct {
-				Name     string   `json:"name"`
-				Version  string   `json:"version"`
-				Licenses []string `json:"licenses"`
-				CPEs     []string `json:"cpes"`
-				PURL     string   `json:"purl"`
-			}
-
-			var syftOutput struct {
-				Artifacts []artifact `json:"artifacts"`
-				Source    struct {
-					Type   string `json:"type"`
-					Target string `json:"target"`
-				} `json:"source"`
-				Schema struct {
-					Version string `json:"version"`
-				} `json:"schema"`
-			}
+			var syftOutput syftOutput
 
 			err = json.Unmarshal(syft.Bytes(), &syftOutput)
 			Expect(err).NotTo(HaveOccurred(), syft.String())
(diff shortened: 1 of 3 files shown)


sbom/sbom_test.go
```diff
@@ -48,56 +47,36 @@ func testSBOM(t *testing.T, context spec.G, it spec.S) {
...
 			}
 
 			var syftOutput struct {
 				Artifacts []artifact `json:"artifacts"`
+				Source    struct {
+					Type   string `json:"type"`
+					Target string `json:"target"`
+				} `json:"source"`
+				Schema struct {
+					Version string `json:"version"`
```
sophiewigmore: In https://github.com/paketo-buildpacks/packit/pull/271/files, I noticed you added a `Schema.URL` field as well. Should we include that here too for completeness?

sbom/sbom_test.go
```diff
@@ -107,51 +86,47 @@ func testSBOM(t *testing.T, context spec.G, it spec.S) {
...
+				BOMFormat   string `json:"bomFormat"`
+				SpecVersion string `json:"specVersion"`
+				Metadata    struct {
+					Component struct {
+						Type string `json:"type"`
+						Name string `json:"name"`
+					} `json:"component"`
 				} `json:"metadata"`
+				Components []component `json:"components"`
 			}
```
sophiewigmore: Can we share the output types between this test and formatted reader since they're the same? Same question  goes for the other formats too.

sophiewigmore (changes requested):

sbom/sbom_test.go
fg-j: I don't think the SBOM URL is very useful to assert on, since it's just a pointer to the schema definition. I think asserting on the schema version is useful, though, because we'll probably want to know if syft starts producing a non-backwards-compatible BOM. See [the README](https://github.com/anchore/syft/blob/a18fbacae7bd375a38e6ece836b6c85571deb80f/schema/json/README.md) of the Syft schema directory for an explanation of "SchemaVer", a cousin of SemVer.

sbom/sbom_test.go
fg-j: Done in the most recent commit.

sophiewigmore:
@fg-j it looks like a couple structs got left behind and are causing linting errors. Once that's fixed this PR is good!

sophiewigmore (approved):

PR #276, opened 2022-01-08
fix(servicebindings): trim provider/type whitespace

### LINKED ISSUE #275: servicebindings.Resolve(...) fails if provider/type files contain whitespace
## What happened?

Create a service binding that can be resolved by `type` and `provider` using `servicebindings.Resolve(...)`

Given the following method of creating the binding, `servicebindings.Resolve("some-type", "some-provider", "")` should succeed:

```bash
mkdir my-binding
echo "some-type" > my-binding/type
echo "some-provider" > my-binding/provider
echo "value" > my-binding/value
```

The binding failed to resolve because `echo` and other means of creating files will append `\n`:

```bash
$ < my-binding/type | od -c
0000000    s   o   m   e   -   t   y   p   e  \n
0000012
$ < my-binding/provider | od -c
0000000    s   o   m   e   -   p   r   o   v   i   d   e   r  \n
0000016
```

Since the `loadBinding` logic does not consider whitespace and [only does a case insensitive equals](https://github.com/paketo-buildpacks/packit/blob/b0348e1e02527ce297517a95def39ac1a4f72fcc/servicebindings/resolver.go#L65), the binding fails to load. It shouldn't be expected that all `type` and `provider` files are free from newlines [since the POSIX standard defines a line in this manner](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206):

> A sequence of zero or more non- `<newline>` characters **plus a terminating `<newline>` character**.

This may also lead to user confusion as they may accidentally create the `type` and `provider` files with a newline and not realize why buildpacks are failing to use their bindings - especially if the buildpacks silently ignore missing optional bindings.

## Build Configuration

N/A - calling `packit` directly via code

## Checklist

noamichael:
## Summary

Closes #275 

Currently, the `servicebindings.Resolve(...)` function does not trim
whitespace when comparing the requested type/providers with what's on the
filesystem. This change ensures that the type/providers read from disk
are sanitized before they are used.

## Checklist

### DIFF
--- a/servicebindings/resolver.go
+++ b/servicebindings/resolver.go
@@ -165,6 +165,7 @@ func loadBinding(bindingRoot, name string) (Binding, error) {
 	if err != nil {
 		return Binding{}, err
 	}
+	binding.Type = strings.TrimSpace(binding.Type)
 	delete(entries, "type")
 
 	provider, ok := entries["provider"]
@@ -173,6 +174,7 @@ func loadBinding(bindingRoot, name string) (Binding, error) {
 		if err != nil {
 			return Binding{}, err
 		}
+		binding.Provider = strings.TrimSpace(binding.Provider)
 		delete(entries, "provider")
 	}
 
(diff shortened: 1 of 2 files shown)


ryanmoran (approved):
This looks great. Thanks!

PR #280, opened 2022-01-19
Make LeveledLogger's writers public 

fg-j:
## Summary

While working on https://github.com/paketo-buildpacks/dotnet-publish/issues/275, I ran into a snag: the `io.Writer`s on the `LeveledLogger` are private, and thus can't be used for streaming logs from command executions. 

As a proof of concept, made the writers public so that I could do [something like this](https://github.com/paketo-buildpacks/dotnet-publish/blob/b0a115d52ea737bf5c6bc472d535d045b80622ae/dotnet_publish_process.go#L61-L70) to stream the logs when the log level is `DEBUG`.

@paketo-buildpacks/tooling-maintainers, my idea isn't fully baked yet. I have a couple questions:
-  Are there drawbacks to making the writers public?
- As written, the code in my [proposed change to dotnet publish](https://github.com/paketo-buildpacks/dotnet-publish/blob/b0a115d52ea737bf5c6bc472d535d045b80622ae/dotnet_publish_process.go#L61-L70) double-prints the output of `dotnet publish` if the command fails and the log level is `DEBUG` (see screenshot below).  I can't think of a way around this without checking the value of `BP_LOG_LEVEL` [here](https://github.com/paketo-buildpacks/dotnet-publish/blob/b0a115d52ea737bf5c6bc472d535d045b80622ae/dotnet_publish_process.go#72), which feels clunky. Does anyone see another way?

In general, I think streaming logs in debug mode will be useful in many buildpacks, and I wanted to start the conversation about how we'll design a solution.

image

## Checklist


fg-j:
@paketo-buildpacks/tooling-maintainers From what I understand, there's more to discuss here than this API change. One of the relevant topics under discussion is whether the buildpack should stream the output of commands like `dotnet publish` _all the time_, not just when `DEBUG` is set. I'm going to bring up this topic at the Working Group meeting tomorrow.

fg-j:
We discussed this topic at a working group meeting and it resulted in a change to the [Paketo Buildpacks Style Guide](https://github.com/paketo-buildpacks/community/blob/c1fb2014df7ad102fd7206d21ba584bcb7651da7/STYLEGUIDE.md), which now states:

> If a buildpack invokes a tool that produces logs (e.g. npm install), it is preferred that the buildpack stream the logs (rather than buffering/swallowing them) to provide maximum visibility to users. The buildpack may adjust the verbosity of a tool's logs or apply formatting to them to improve the overall build logs' readability.

This captures the sentiment of buildpack maintainers that streaming logs is preferable as log as the resulting bulid output is readable overall.

So I've tweaked my approach in the dotnet-publish buildpack to simply [stream the output of `dotnet publish` in all cases](https://github.com/paketo-buildpacks/dotnet-publish/blob/55f7e8e5c4ed5328b1e7fc591986bd668cd023df/dotnet_publish_process.go).

I chose to expose`scribe.Emitter.ActionWriter` so that the log output from `dotnet publish` can be formatted in a way that's consistent with other buildpack output, without breaking the current log formatting abstraction. To achieve consistent formatting using the `scribe.Writer` directly, I'd need to dig into the implementation  of `scribe.Emitter.Action()`  to see how the logs are formatted. I think it's better to abstract log formatting away from buildpack authors as much as possible.

PR #289, opened 2022-02-23
Fixes error interpolation case where err is nil

ryanmoran:
## Summary

In cases where the checksums for dependencies don't match, we were outputting an error message that looked broken:

```
[builder]   Executing build process
[builder]     Installing Node Engine 16.13.2
[builder] checksum does not match: %!s(<nil>)
```

This was caused by our attempt to interpolate `nil` as a string in the error message. We don't need this interpolation because we have already checked that the error is `nil` in prior statements.

## Checklist

### DIFF
--- a/postal/service_test.go
+++ b/postal/service_test.go
@@ -616,7 +616,7 @@ version = "this is super not semver"
 						"",
 					)
 
-					Expect(err).To(MatchError(ContainSubstring("checksum does not match")))
+					Expect(err).To(MatchError("validation error: checksum does not match"))
 				})
 			})
 
--- a/postal/service_test.go
+++ b/postal/service_test.go
@@ -572,6 +572,17 @@ version = "this is super not semver"
 		})
 
 		context("failure cases", func() {
+			context("when dependency mapping resolver fails", func() {
+				it.Before(func() {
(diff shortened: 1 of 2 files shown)


postal/service_test.go
```diff
@@ -616,7 +616,7 @@ version = "this is super not semver"
 						"",
 					)
 
-					Expect(err).To(MatchError(ContainSubstring("checksum does not match")))
```
sophiewigmore, line 619: This test is testing https://github.com/paketo-buildpacks/packit/blob/86f0343bcc395bdbc4f3a569d54e6df317635faa/postal/service.go#L172-L175 I think. Is that intended?

ryanmoran: Yeah. I missed that we had an earlier test that asserts we handle the error coming back from the decompression. Unfortunately, I've been unable to create a test case that covers what I really want to test, which is the `validatedReader.Valid()` conditional. I can't seem to create a tarball that triggers the situation I am able to see in a real dependency in a real buildpack.

ForestEckhardt: What buildpacks are you seeing this error in?

ForestEckhardt: I have constructed a buffer that can get to the second failure state. https://gist.github.com/ForestEckhardt/84eb2305c4626e3b93a5457baed20764

sophiewigmore (approved):

PR #300, opened 2022-03-15
Automatically move listed `Exec.D` executables into `<layer>/exec.d/<executable>`

joshuatcasey:
## Summary
Automatically move listed `Exec.D` executables into `<layer>/exec.d/<executable>`

## Use Cases
This allows buildpacks such as `dotnet-execute` to have logic such as this...

```
portChooserLayer, err := context.Layers.Get("port-chooser")
if err != nil {
	return packit.BuildResult{}, err
}

portChooserLayer.Launch = true
portChooserLayer.ExecD = []string{filepath.Join(context.CNBPath, "bin", "port-chooser")}

return packit.BuildResult{
	Layers: []packit.Layer{
		portChooserLayer,
	},
		Processes: processes,
	},
}, nil
```

... instead of custom logic such as this:
- https://github.com/paketo-buildpacks/dotnet-execute/blob/91eed322ffba826d64f7428c8ca7b69839cfbfb7/build.go#L95-L112
- https://github.com/paketo-buildpacks/node-engine/blob/00b9d1902d76bbafcea8a33c8a80e9b73e24273d/environment.go#L39-L48

## Checklist


ryanmoran:
I'm wondering if it might be more fluent of an API if it looked something like the following:

```go
portChooserLayer, err := context.Layers.Get("port-chooser")
if err != nil {
	return packit.BuildResult{}, err
}

portChooserLayer.Launch = true
portChooserLayer.ExecD = []string{filepath.Join(context.CNBPath, "bin", "port-chooser")}

return packit.BuildResult{
	Layers: []packit.Layer{
		portChooserLayer,
	},
		Processes: processes,
	},
}, nil
```

You'd then have the top-level `packit.Build` function handle the work that was done in `commands.PopulateExecDCommands` like we do for other layer-related concerns: https://github.com/paketo-buildpacks/packit/blob/ed29b6310903d939b4cba558c14c68d548d5b368/build.go#L189-L236

commands/populate_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+		context("failure cases", func() {
+			it("throws a specific error when command executable not found", func() {
+				_, err := production.PopulateExecDCommands(packit.BuildContext{
+					CNBPath: cnbDir,
+					Layers: packit.Layers{
+						Path: layersDir,
+					},
+				}, layerName, "command-that-is-not-found")
+				Expect(err).To(MatchError("file bin/command-that-is-not-found does not exist. Be sure to include it in the buildpack.yml"))
```
fg-j:
```suggestion
				Expect(err).To(MatchError("file bin/command-that-is-not-found does not exist. Be sure to include it in the buildpack.toml"))
```

commands/populate.go
```diff
@@ -0,0 +1,64 @@
...
+	err = os.MkdirAll(execdDir, os.ModePerm)
+	if err != nil {
+		return layer, err
+	}
+
+	lexicalWidth := 1 + int(math.Log10(float64(len(commands))))
+
+	for i, command := range commands {
+		filename := filepath.Join("bin", command)
+		source := filepath.Join(context.CNBPath, filename)
```
fg-j: I'm wondering if it would be useful to make this more configurable -- i.e. accept the directory/directories in which to look for executables as an input, rather than assuming that they'll always be in `context.CNBPath/bin`. This could enable a buildpack to create or install an executable at build time and then still set it up as an `exec.d`

fg-j:
All in all, I think this will be a very useful addition as buildpack authors are forced to move away from `.profile` scripts / shell execution.

commands/populate.go
joshuatcasey: That's a good call. @ryanmoran 's suggestion above had a similar suggestion, so it sounds like Paketo shouldn't require `/bin` as a path prefix. 

I don't have a solution in mind right now, since the API will likely change (again as per @ryanmoran 's suggestion).

commands/populate_test.go
joshuatcasey: D'oh

joshuatcasey:
@ryanmoran I like your suggestion. I'm curious if we can go a step further and eliminate the explicit layer creation. Can `packit.Build` just make a layer? That could look something like so:

```
return packit.BuildResult{
        ExecD: []string{
                filepath.Join(context.CNBPath, "bin", "port-chooser"),
                filepath.Join(context.CNBPath, "bin", "other-exec-d-cmd"),
                filepath.Join("/foo", "bar", "something-else"),
        },
		Processes: processes,
	},
}, nil
```

ryanmoran:
> Can `packit.Build` just make a layer?

I see a couple concerns with that.

1) Its more magical than I think we would expect compared to the other APIs in `packit`.

2) It doesn't allow buildpack authors to separate their exec.d processes based on how their layers get added to the launch image. An example might be a buildpack that offers 2 different dependencies with 2 different exec.d processes. In that case, I might want to make it so that each one could be added as a build or launch layer exec.d process independently.

joshuatcasey:
@ryanmoran could you take a look?

build.go
```diff
@@ -233,6 +235,35 @@ func Build(f BuildFunc, options ...Option) {
...
+			lexicalWidth := 1 + int(math.Log10(float64(len(layer.ExecD))))
+
+			for i, exe := range layer.ExecD {
+				if exists, err := fs.Exists(exe); !exists {
+					config.exitHandler.Error(fmt.Errorf("file %s does not exist. Be sure to include it in the buildpack.toml", exe))
+					return
+				} else if err != nil {
+					config.exitHandler.Error(err)
+					return
+				}
```
ryanmoran: Do we need this explicit existence check when `fs.Copy` should fail with the same class of error saying that the file does not exist?

build_test.go
```diff
@@ -1079,6 +1081,156 @@ api = "0.5"
...
+				Expect(exists).To(Equal(true), fmt.Sprintf("file %s not found", fullName))
+				Expect(err).NotTo(HaveOccurred())
+			}
+
+			checkNotExists = func() {
+				fullName := filepath.Join(layersDir, "layer-with-exec-d-stuff", "exec.d")
+				exists, err := fs.Exists(fullName)
+				Expect(exists).To(Equal(false), fmt.Sprintf("file %s should not be found", fullName))
+				Expect(err).NotTo(HaveOccurred())
+			}
```
ryanmoran: I'm pretty sure you can get the same coverage using the `gomega` builtin `BeARegularFile`. You'd have something like

```go
Expect(filepath.Join(layersDir, "layer-with-exec-d-stuff", "exec.c", filepath.Base(exe0))).To(BeARegularFile())
```

joshuatcasey: Thanks for the tip! Will pull that into a few places here and elsewhere.

build.go
joshuatcasey: I wanted to explicitly call out adding the file to `buildpack.toml` since that's an additional manual step on top of creating `cmd/new-command/main.go`. If it doesn't add value I'm happy to remove the additional branch.

ryanmoran: If that's what you want, you can check the error coming back from `fs.Copy` to see if it is `os.ErrNotExist`, something like the following:

```go
err = fs.Copy(exe, destination)
if err != nil {
  if errors.Is(os.ErrNotExist) {
    err = fmt.Errorf("file %s does not exist. Be sure to include it in the buildpack.toml", exe)
  }

  config.exitHandler.Error(err)
  return
}
```

joshuatcasey: done, good catch

ryanmoran (approved):

fg-j (approved):

Issue #302, opened 2022-03-16
Support varying schema versions of SBOM outputs


ryanmoran:
## Problem
Relying upon the SBOM formatting code available in `syft` is creating inadvertent breaking changes in our API contract. For example, we cannot currently upgrade the version of `syft` that we are using without also changing the schema version of our CycloneDX SBOM output.

In order to stabilize the contract in the long run, we should be introduce functionality in `packit/sbom` capable of generating SBOM format outputs in many different schema versions of each SBOM type. Doing so will allow us to keep the SBOM output, and our API, stable while still consuming the latest versions of the `syft` library for scanning and generating the SBOM data.

## Proposal
Our current implementation leverages the [`syft.Encode`](https://pkg.go.dev/github.com/anchore/syft/syft@v0.41.6#Encode) method to convert a [`sbom.SBOM`](https://pkg.go.dev/github.com/anchore/syft@v0.41.6/syft/sbom#SBOM) into `[]bytes` that can be written to a file.

https://github.com/paketo-buildpacks/packit/blob/ed29b6310903d939b4cba558c14c68d548d5b368/sbom/formatted_reader.go#L46-L49

The `packit/sbom` package does not currently implement any of the formatting itself. Instead, it leverages the existing formats made available by `syft`:

https://github.com/paketo-buildpacks/packit/blob/ed29b6310903d939b4cba558c14c68d548d5b368/sbom/formatted_reader.go#L34-L44

Unfortunately, these formats are updated on a near-constant basis, and in a way that breaks backwards-compatibility in the SBOM output format contract.

Luckily, the `syft` library has recently undergone a refactoring that now allows us to define our own format types that can be used to encode SBOM output. To implement a format, you need to create a concrete type that conforms to the [`sbom.Format` interface](https://pkg.go.dev/github.com/anchore/syft@v0.41.6/syft/sbom#Format)

```go
type Format interface {
	ID() FormatID
	Encode(io.Writer, SBOM) error
	Decode(io.Reader) (*SBOM, error)
	Validate(io.Reader) error
}
```

We should implement a package internal to `packit/sbom` that implements at least the following formats (all versions we have ever released support for through the existing `sbom` package):

* [x] CycloneDX JSON 1.3
* [ ] CycloneDX JSON 1.4
* [ ] Syft 2.0.0
* [ ] Syft 2.0.1
* [x] Syft 2.0.2
* [ ] Syft 3.0.0
* [x] Syft 3.0.1
[...]

PR #312, opened 2022-03-24
Implement Syft2, 3 and CycloneDX 1.3 SBOM formats

fg-j:
## Summary

This PR pulls in *a lot* of borrowed code from anchore/syft and CycloneDX/cyclonedx-go, because both of these libraries do not currently support multiple versions of their SBOM schema. I've had to yank sections of old versions of these codebases and embed them here.

As much as possible, I've attempted to document the sources of the code – and where things diverge from their original content.

It's my intent not to change the API of packit's SBOM package at all with this PR, given how experimental this featureset is. The hope is that if this is added to packit and [this change to the lifecycle](https://github.com/buildpacks/lifecycle/issues/828) is released, Paketo buildpacks can pin to CycloneDX 1.3 SBOMs.

Supercedes #325  as well.

Partially addresses #302 , starting with requested / backward-incompatible SBOM formats. Ideally, I'd like to push these implementations of the Syft sbom.Format interface into syft itself. Our use of them in packit can serve as a proof of concept of the viability of supporting multiple SBOM formats.

## Checklist


fg-j:
There are some TODOs in here for tweaking parts of the implementation. Would love feedback on approaching those! I've been staring at it all too long. Need some fresh eyes to help me tighten things up.

fg-j:
Ugh. The linter is failing

fg-j:
Merging this is blocked on the release of [the corresponding lifecycle feature](https://github.com/buildpacks/lifecycle/issues/828) -- since we'll need to use the feature to update the `sbom-formats` array in `buildpack.toml`s to pin to CycloneDX 1.3 SBOM.

sbom/sbom_format.go
```diff
@@ -0,0 +1,101 @@
...
+		selected = spdxFormats[version]
+	case SyftFormat:
+		selected = syftFormats[version]
+	default:
+		return sbomFormat{}, fmt.Errorf("unsupported SBOM format: '%s'", mediaType)
+	}
+
+	if selected == sbom.FormatID("") {
+		return sbomFormat{}, fmt.Errorf("version '%s' is not supported for SBOM format '%s'", version, baseType)
+	}
```
sophiewigmore, line 91: Why is it necessary to check `selected` against `sbom.FormatID("")`? If there's no matching format/version combo, won't selected just be equal to `""`?

sbom/sbom_format.go
```diff
@@ -0,0 +1,101 @@
...
+	return sbomFormatByID(selected)
+}
+
+func sbomFormatByID(id sbom.FormatID) (sbomFormat, error) {
```
sophiewigmore, line 95: I've been looking at this function for a bit now, but I really can't think of a better way to do it than what you've laid out

sophiewigmore:
@fg-j as discussed, it seems like the path of this PR is subject to change based off of our conversation with the Syft team at their working group meeting today. Do you want to pause on reviews until we have a clearer picture of what will happen in Syft first?

ryanmoran:
I think we'll still want to merge this PR or something like it in the short-term so that we can unblock other issues, but we have a much higher degree of confidence that this debt we will be incurring can be paid down at some point in the medium-term future. That being said, I think the goals for what we should try to support here are to lock down the formats we currently support as outputs so that we can bump the Syft library forward to the latest versions without breaking our API contract.

Specifically, we need support for:
* [x] CycloneDX 1.3
* [x] Syft 3.0.1 (used in Buildpacks)
* [x] Syft 2.0.2 (used in Stacks)
* [x] SPDX 2.2

We get Syft 3 and SPDX 2.2 for now with what is on the latest versions of Syft, but there is the concern that it might be broken at some point in the future. Should we duplicate support in our own package to make sure the API stays stable, or should we just keep up with checking that the schema versions don't change for those formats?

fg-j:
@ryanmoran Given that syft has shown some interest in exposing APIs that'll make it easier to create SBOM format implementations (without copying in their internal packages), I'm inclined to do the minimum (i.e. add Syft 3.0.1) in the short term and wait to see if we can unload some tech debt before we add our own CycloneDX 1.4 and SPDX 2.2 implementations.

I will update this PR to make CycloneDX 1.3  the default SBOM version  generated when no `version` parameter is set on the media type so that consuming buildpacks will stay pinned to a CycloneDX minor version.

sbom/sbom_format.go
fg-j: Resolved in latest version.

sophiewigmore:
@fg-j you mentioned that when new Syft versions come out we'll occasionally have to change our implementation like in https://github.com/paketo-buildpacks/packit/pull/312/commits/7335806a473391a5703ff51447a2a9c966de4e80. Once this is merged in, what kind of Syft changes do we need to be cognizant of in releases? I'd imagine it will differ from what we're on the lookout for now (backwards incompatible schema changes).

fg-j:
@sophiewigmore  – For CycloneDX 1.4 and SPDX 2.2, since this PR still relies on Syft's implementations of some schema versions, we'll still want to assert that the SBOMs generated with the syft Formats have the expected versions.

We test this [here](https://github.com/paketo-buildpacks/packit/blob/6968f3cb09b32a74fb50c531da478864b493004b/sbom/sbom_test.go#L220) for CycloneDX 1.4 and [here](https://github.com/paketo-buildpacks/packit/blob/6968f3cb09b32a74fb50c531da478864b493004b/sbom/sbom_test.go#L155) for SPDX 2.2

When syft changes the underlying structs it uses to represent packages, we may need to change the code that converts from the syft representation to the JSON output representations. The entrypoint for this logic is in the `to_format_model.go` file in each format subdirectory (e.g. in [syft 3.0.1](https://github.com/paketo-buildpacks/packit/blob/6968f3cb09b32a74fb50c531da478864b493004b/sbom/internal/formats/syft301/to_format_model.go)). The `encode_test.go` for each format should fail if the syft structs change in a way that makes them convert differently into JSON.

If syft changes its internal representation of packages significantly, we may need to write bespoke converters, like [`ConvertImageMetadata`](https://github.com/paketo-buildpacks/packit/blob/6968f3cb09b32a74fb50c531da478864b493004b/sbom/internal/formats/syft2/source/image_metadata.go#L5), which peels fields off syft's current `source.ImageMetadata` type and returns them in a struct that matches the legacy representation of the same data.

fg-j:
Removing the blocked label because this implementation defaults to producing CycloneDX 1.3 and Syft 3.0.1 SBOMs, which are the same formats we currently generate. Merging this will **not** change the schema of buildpack generated SBOMs. No longer blocked on the release of [the lifecycle change allowing schema versions](https://github.com/buildpacks/lifecycle/issues/828).

sophiewigmore:
@fg-j I've gone through the code a few times and it looks good to me! It makes sense and I think covers all of our bases. @ryanmoran tagging you for a review as well before we merge it in to make sure it aligns with what you expected in #302 / to provide more technical expertise on the code structure.

I've also tried it out with the defaults as you mentioned, and the output looks as expected.

sophiewigmore (dismissed):

ryanmoran:
I'm trying to get `npm-install` to run with this branch. I'm having some trouble understanding how I can specify what format I want for a `sbom.FormattedReader`. For example, how do I specify the right formats for these test values?

https://github.com/paketo-buildpacks/npm-install/blob/28857f234fee9c4bbbc67cf4fa4117a4bfe776c2/build_test.go#L160-L173

fg-j:
@ryanmoran The latest commit here should fix those test failures -- the `NewFormattedReader()` now does more to convert the inputted `packit.sbom.Format` into a `syft.sbom.Format`, respecting the schema version defaults we've set in packit.

ryanmoran (approved):

PR #331, opened 2022-04-16
[DRAFT] Using generics to add planning metadata

joshuatcasey:
## Summary
Just exploring the use of generics

## Use Cases
Helps a little with type safety. See paketo-buildpacks/dotnet-core-runtime#325 for a possible use of this.

## Checklist


joshuatcasey:
Sorry for the notification noise on this one - I accidentally created a non-draft PR :(

robdimsdale:
In principle I like the direction this is going.

One issue of this current implementation is that the design philosophy of `packit` is that the top-level package represents the implementation of the buildpack spec - nothing more; nothing less - and therefore it should not depend on any sub packages. So we would have to lift some or all of the contents of the `planning` package up to the top-level package.

Additionally, I haven't convinced myself that I like the naming scheme (e.g. `DetectFuncOf[]`, `BuildResultOf[]`). On the one hand it reads ok - "a build result of type T" - but on the other hand it feels verbose and I don't like that callers have to refer to the type each time (e.g. `packit.DetectResultOf[planning.Metadata]`). Is there a way we can use type-definition to expose a typed function with this generic type hard-coded? I haven't played enough with golang generics to know if this is possible, but if possible I would like to see something like `type DetectResult DetectResultOf[planning.Metadata]`. This would encapsulate the generics within packit, which seems reasonable given packit does not support any alternative values of `T` for `DetectResultOf[T]` and provides a cleaner interface for the consumer.

joshuatcasey:
I think what you mention is possible, but  it broke backwards-compatibility for consumers.

- Better but not backwards-compatible: `type BuildPlan = BuildPlanOf[planning.Metadata]`
- Less nice but backwards-compatible: `type BuildPlan = BuildPlanOf[interface{}]`

IIRC my thought process around types in `planning` vs root Packit was that this was just an optional type - other consumers of packit could have their own `BuildPlanOf[MyCustomType]`. So while `planning` types were the default, they weren't technically required. But that's a bit shaky logic.

To be honest I wonder if we could pull the Paketo metadata up to the spec level. It feels weird that many of our buildpacks impose a structure on top of what's in the spec (as part of architecture and not convenience), and each buildpack seems to have a slightly different implementation for this. Between that variation and most of the Paketo READMEs suggesting that `[requires.version]` is respected it took me a while to understand the build plan.

For example, I think this README is wrong when it suggests that `requires.version` is respected: https://github.com/paketo-buildpacks/miniconda. `requires.metadata.version` should be respected.

robdimsdale:
Gotcha, yeah that makes sense about the backwards-compatibility. In principle, I general prefer to go for the cleanest API even if it means a breaking change.

However, the more I think about this concept of generics, the more I think that it's an (elegant) solution looking for a problem.

Specifically, I think that we would be better suited to having a single, concrete implementation of the commonly-used metadata in a subpackage of packit, possibly with smaller libraries for functionality that is only common to a subset of buildpacks, rather than aiming for a generic solution like you've outlined. I think this for a few reasons:

1. The CNB spec has no structure for the metadata, so as long as the top-level `packit` package aims to map 1:1 onto the spec, it will always have to have `map[string]interface{}`. We'd have to manually erase all type information when converting to the top-level packit interface.
2. Packit is design to be an opinionated library for facilitating shared behavior across the various paketo buildpacks. There is only one implementation for the metadata, which is fields common to most buildpacks (like `VersionSource`). I can't see a world where we have two commonly-used implementations such that generics would make sense. And even if we did, we would just have two concrete implementations without leveraging generics, because we can't use generics at the top-level anway (see above).
    - There are fields that are only common to a subset of buildpack interactions (like `dependency-sha`) - these probably don't belong in packit anyway, instead they might be better served in some smaller library consumed only by the relevant buildpacks (e.g. one shared library for the python buildpack interactions, another shared library for the php interactions, etc).

Even if the CNB spec changed to have specific fields in the metadata alongside a general map, I still don't think generics would be possible for the same reasons above. Practically, if such a change happened then we would encode those known fields in the top-level `packit` package alongside a `map[string]interface{}` and we would have essentially the same problem above.

joshuatcasey:
Since this PR was really only for discussion, I'll close it now.

PR #335, opened 2022-04-21
API 0.8: Uses environment variables over postional args when available

### LINKED ISSUE #322: Implement Buildpack RFC 0100: Replace positional args to buildpack executables with env vars
## Blocked
~~This issue is blocked on the release of Buildpack API 0.8.~~

## Summary
Buildpack [RFC0100](https://github.com/buildpacks/rfcs/blob/main/text/0100-buildpack-input-vars.md) replaces positional arguments to build and detect executables with named environment variables as of Buildpack API 0.8 and beyond.

For detection, rather than passing in two [positional arguments](https://github.com/paketo-buildpacks/packit/blob/1c5ad2f17e60d9988d4e9e1518207c54fe8dd66b/detect.go#L89) `<platform>` and `<plan>`, environment variables `CNB_PLATFORM_DIR` and `CNB_BUILD_PLAN_PATH` can be used respectively.

For build, rather than [passing in three positional arguments](https://github.com/paketo-buildpacks/packit/blob/1c5ad2f17e60d9988d4e9e1518207c54fe8dd66b/build.go#L99-L103) for `<layers>`, `<platform>`, and `<plan>` environment variables `CNB_LAYERS_DIR`, `CNB_PLATFORM_DIR`, and `CNB_BP_PLAN_PATH` can be used respectively.

`CNB_PLATFORM_DIR` is the absolute path to the platform directory.
`CNB_BUILD_PLAN_PATH` is the absolute path to the build plan
`CNB_LAYERS_DIR` is the absolute path to the buildpack layers directory
`CNB_BP_PLAN_PATH` is the relevant [Buildpack Plan entries](https://github.com/buildpacks/rfcs/blob/main/text/0100-buildpack-input-vars.md#buildpack-plan-toml) from detection (TOML)

## Proposal
If the buildpack is using Buildpack API 0.8 or higher, the above arguments can be provided with the [same mechanism](https://github.com/paketo-buildpacks/packit/blob/1c5ad2f17e60d9988d4e9e1518207c54fe8dd66b/build.go#L118) used to provide `CNB_BUILDPACK_DIR` in both packit's detec and build. For each of the relevant fields, a `LookupEnv` should be performed to incorporate that value if it is set.

ForestEckhardt:
Resolves #322 

## Checklist

### DIFF
--- a/build.go
+++ b/build.go
@@ -3,12 +3,13 @@ package packit
 import (
 	"errors"
 	"fmt"
-	"github.com/paketo-buildpacks/packit/v2/fs"
 	"math"
 	"os"
 	"path/filepath"
 	"sort"
 	"strings"
+
+	"github.com/paketo-buildpacks/packit/v2/fs"
 
 	"github.com/BurntSushi/toml"
 	"github.com/Masterminds/semver/v3"
@@ -96,16 +97,15 @@ func Build(f BuildFunc, options ...Option) {
 		config = option(config)
 	}
 
-	var (
-		layersPath   = config.args[1]
-		platformPath = config.args[2]
-		planPath     = config.args[3]
-	)
-
 	pwd, err := os.Getwd()
 	if err != nil {
 		config.exitHandler.Error(err)
 		return
+	}
+
+	planPath, ok := os.LookupEnv("CNB_BP_PLAN_PATH")
+	if !ok {
+		planPath = config.args[3]
 	}
 
 	var plan BuildpackPlan
@@ -118,6 +118,16 @@ func Build(f BuildFunc, options ...Option) {
 	cnbPath, ok := os.LookupEnv("CNB_BUILDPACK_DIR")
 	if !ok {
 		cnbPath = filepath.Clean(strings.TrimSuffix(config.args[0], filepath.Join("bin", "build")))
+	}
+
+	layersPath, ok := os.LookupEnv("CNB_LAYERS_DIR")
+	if !ok {
+		layersPath = config.args[1]
+	}
+
+	platformPath, ok := os.LookupEnv("CNB_PLATFORM_DIR")
+	if !ok {
+		platformPath = config.args[2]
 	}
 
 	var buildpackInfo struct {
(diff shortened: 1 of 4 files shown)


sophiewigmore:
@ForestEckhardt I'm good with your decision not to use an environment variable library beyond `os`, it seems like overkill to use one in this case. 

One question - I noticed that the code falls back to the positional arguments if the environment variable is not set, getting around the need for an API version check. However, if the positional arguments are fully removed, I'm wondering if we'd want to throw an error if the environment variable isn't set? 
Also, what would we expect the behaviour should be in API 0.7 and below if the environment variables are somehow set? I think we'd still want to use the positional arguments, but correct me if I'm wrong!

ForestEckhardt:
@sophiewigmore I did this fall back for a couple of reasons.
1. I am actually unsure when/if they are removing the positional args. I figure going back to the old style is safe enough to allow other platforms to catch up.
2. If neither the env var or the positional arg is set then we are using a non-standard platform and there are going to be issues regardless. I expect one of them to be set.

sophiewigmore:
@ForestEckhardt to point 1, that is a fair assessment. If they do remove positional arguments, then things may break and we'd have to back in and fix them later. It's just a matter of preference because your way will work as things are now.

To point 2, the env. vars are only considered in API 0.8, so it would be erroneous to let users set them in that case. Is your opinion that we should just leave it to the lifecycle to fail if a buildpack author does something wrong? Or should we enforce the constraints up front?

ForestEckhardt:
@sophiewigmore My point for 2 is that I don't understand why we should bother checking the API version. The information will be given to us either via the env vars or the positional args. For me it makes no functional difference where I get the information from. In fact I would hope that platforms for some amount of time support both setting the env var and positional args to support multiple buildpack API versions, which is what I am assuming that platforms are doing at the moment. If this were grossly new functionality (such as the process specific working dir) I would be more concerned but in this case I am not to worried.

To put a fine point on it I think that erroring just because the env var is set is not a good idea because I imagine that it would still be set by the platform regardless. I could write some logic to only check the env vars given the API version is 0.8 or greater but I think that is generally clunky and unnecessary, especially because this information that is passed to the buildpacks and not from the buildpacks to the lifecycle.

sophiewigmore:
@ForestEckhardt you have me convinced, I think comparing this against "grossly new functionality" helped me understand what you mean that it doesn't need to be so strict in this specific case

sophiewigmore (approved):

PR #342, opened 2022-05-06
Add sbom.Autobom func to magic away some SBOM boilerplate

joshuatcasey:
Add `sbom.Autobom` func to magic away some boilerplate around SBOM generation

⚠️ This is mostly here to prompt some discussion. 
- Do we need this magic?
- Should we put this magic into the existing `sbom.Generate` func?
- Should we do something different with this magic? Perhaps we just tell the `Build` func that we want an SBOM and it does the magic.

See paketo-buildpacks/poetry-install#37 for an example of how this might simplify buildpack code.

## Checklist

### DIFF
--- a/sbom/autobom.go
+++ b/sbom/autobom.go
@@ -1,0 +1,36 @@
+package sbom
+
+import (
+	"time"
+
+	"github.com/paketo-buildpacks/packit/v2"
+	"github.com/paketo-buildpacks/packit/v2/chronos"
+	"github.com/paketo-buildpacks/packit/v2/scribe"
+)
+
+// Autobom will return a packit.SBOMFormatter containing the SBOM for the given path.
+// It will also log information about the SBOM generation, including how long it took.
+func Autobom(
+	path string,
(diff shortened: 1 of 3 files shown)


ryanmoran:
To me, the pattern that is being distilled here is very specific to the buildpacks that we maintain directly in the project. However, `packit` is meant to be more general-purpose. We have packages that make choices, but you can use them somewhat independently from one another. In this case, we've merged together several choices in a way that prevents decomposition. You may have noticed that before this PR, we've never had logging in any part of `packit` outside of `scribe`. The choice here has always been that we want to be flexible enough that users could choose something different for the constituent parts of `packit`.

Now, I think I can buy that this reduces repetition in some areas. But, I'm not sure `packit` is really the right place. I could maybe see another library that codifies these types of "workflows" that are common in our buildpacks and provides implementations that can easily be reused.

joshuatcasey:
Makes sense @ryanmoran . There's not much pressure to pull this in so I'll close.

PR #343, opened 2022-05-08
Reduce Watchexec boilerplate and standardize watchexec processes

joshuatcasey:
This PR is meant to reduce boilerplate and standardize how to invoke a `watchexec`-flavored process.

There are a handful of buildpacks that all have the same logic to pull in `watchexec`. This means in order for a buildpack to `require watchexec`, there are often multiple snippets of code that have to be used in the right way.

See paketo-buildpacks/go-build#310 for how this could be used.

Questions:
- Is it worth hiding the implementation to reduce boilerplate?
- Should we use a fakeable func instead of a global func? This could help reduce testing boilerplate in consuming buildpacks.
- Should this logic really exist in the `watchexec` buidpack? Could make some sense for that buildpack to expose some sort of SDK so that other buildpacks could know how to consume it.

### DIFF
--- a/reload/init_test.go
+++ b/reload/init_test.go
@@ -1,0 +1,17 @@
+package reload_test
+
+import (
+	"testing"
+
+	"github.com/onsi/gomega/format"
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+)
+
+func TestReload(t *testing.T) {
+	format.MaxLength = 0
+
+	suite := spec.New("reload", spec.Report(report.Terminal{}))
+	suite("Reload", testReload)
+	suite.Run(t)
+}
--- a/reload/reload.go
+++ b/reload/reload.go
@@ -1,0 +1,92 @@
+package reload
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/paketo-buildpacks/packit/v2"
+)
+
+var LiveReloadEnabledEnvVar = "BP_LIVE_RELOAD_ENABLED"
+var WatchExecRequirement = packit.BuildPlanRequirement{
+	Name: "watchexec",
+	Metadata: map[string]interface{}{
+		"launch": true,
+	}}
+
+func ShouldEnableLiveReload() (bool, error) {
+	if reload, ok := os.LookupEnv(LiveReloadEnabledEnvVar); ok {
+		if shouldEnableReload, err := strconv.ParseBool(reload); err != nil {
+			return false, fmt.Errorf("failed to parse %s value %s: %w", LiveReloadEnabledEnvVar, reload, err)
+		} else if shouldEnableReload {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// ReloadableProcessSpec contains information to be translated directly into watchexec arguments
+type ReloadableProcessSpec struct {
+	// WatchPaths will translate into --watch args
+	// Optional. If len == 0, then no --watch args will be provided
+	WatchPaths []string
+
+	// IgnorePaths will translate into --ignore args
+	// Optional. If len == 0, then no --ignore args will be provided
+	IgnorePaths []string
+
+	// Shell will translate into --shell
+	// Optional. If not provided, will use "none"
+	Shell string
+
+	// VerbosityLevel will translate into -v
+	// If VerbosityLevel is 0, no -v arg will be provided.
+	// If VerbosityLevel is greater than 0, the appropriate number of v's will be added
+	// E.g. VerbosityLevel = 2 => -vv
+	// E.g. VerbosityLevel = 4 => -vvvv
+	VerbosityLevel int
+}
+
+func TransformReloadableProcesses(originalProcess packit.Process, spec ReloadableProcessSpec) (nonReloadable packit.Process, reloadable packit.Process) {
+	nonReloadable = originalProcess
+	nonReloadable.Default = false
+
+	reloadable = originalProcess
+	reloadable.Type = fmt.Sprintf("reload-%s", originalProcess.Type)
+	reloadable.Command = "watchexec"
+	reloadable.Args = buildArgs(originalProcess, spec)
+
+	return nonReloadable, reloadable
+}
+
+func buildArgs(originalProcess packit.Process, spec ReloadableProcessSpec) []string {
+	args := []string{
+		"--restart",
+	}
+
+	for _, watchPath := range spec.WatchPaths {
+		args = append(args, "--watch", watchPath)
+	}
+
+	for _, ignorePath := range spec.IgnorePaths {
+		args = append(args, "--ignore", ignorePath)
+	}
+
+	shell := "none"
+	if spec.Shell != "" {
+		shell = spec.Shell
+	}
+	args = append(args, "--shell", shell)
+
+	if spec.VerbosityLevel > 0 {
+		args = append(args, "-"+strings.Repeat("v", spec.VerbosityLevel))
+	}
+
+	args = append(args,
+		"--",
+		originalProcess.Command)
+	args = append(args, originalProcess.Args...)
+	return args
+}
--- a/reload/reload_test.go
+++ b/reload/reload_test.go
@@ -1,0 +1,393 @@
+package reload_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/paketo-buildpacks/packit/v2"
+	"github.com/paketo-buildpacks/packit/v2/reload"
+	"github.com/sclevine/spec"
+
+	. "github.com/onsi/gomega"
+)
+
+func testReload(t *testing.T, context spec.G, it spec.S) {
+	var Expect = NewWithT(t).Expect
+
+	context("ShouldEnableLiveReload", func() {
+		context("When it should enable live reload", func() {
+			context("when BP_LIVE_RELOAD_ENABLED is 'true'", func() {
+				it.Before(func() {
+					Expect(os.Setenv("BP_LIVE_RELOAD_ENABLED", "true")).To(Succeed())
+				})
+
+				it.After(func() {
+					Expect(os.Unsetenv("BP_LIVE_RELOAD_ENABLED")).To(Succeed())
+				})
+
+				it("should require watchexec at launch", func() {
+					require, err := reload.ShouldEnableLiveReload()
+					Expect(err).NotTo(HaveOccurred())
+					Expect(require).To(Equal(true))
+				})
+			})
+
+			context("when BP_LIVE_RELOAD_ENABLED is '1'", func() {
+				it.Before(func() {
+					Expect(os.Setenv("BP_LIVE_RELOAD_ENABLED", "1")).To(Succeed())
+				})
+
+				it.After(func() {
+					Expect(os.Unsetenv("BP_LIVE_RELOAD_ENABLED")).To(Succeed())
+				})
+
+				it("should require watchexec at launch", func() {
+					require, err := reload.ShouldEnableLiveReload()
+					Expect(err).NotTo(HaveOccurred())
+					Expect(require).To(Equal(true))
+				})
+			})
+
+			context("when BP_LIVE_RELOAD_ENABLED is 'T'", func() {
+				it.Before(func() {
+					Expect(os.Setenv("BP_LIVE_RELOAD_ENABLED", "T")).To(Succeed())
+				})
+
+				it.After(func() {
+					Expect(os.Unsetenv("BP_LIVE_RELOAD_ENABLED")).To(Succeed())
+				})
+
+				it("should require watchexec at launch", func() {
+					require, err := reload.ShouldEnableLiveReload()
+					Expect(err).NotTo(HaveOccurred())
+					Expect(require).To(Equal(true))
+				})
+			})
+		})
+
+		context("when it should not enable live reload", func() {
+			context("when BP_LIVE_RELOAD_ENABLED is not set", func() {
+				it("should not require anything at launch", func() {
+					require, err := reload.ShouldEnableLiveReload()
+					Expect(require).To(Equal(false))
+					Expect(err).NotTo(HaveOccurred())
+				})
+			})
+
+			context("when BP_LIVE_RELOAD_ENABLED is 'F'", func() {
+				it.Before(func() {
+					Expect(os.Setenv("BP_LIVE_RELOAD_ENABLED", "F")).To(Succeed())
+				})
+
+				it.After(func() {
+					Expect(os.Unsetenv("BP_LIVE_RELOAD_ENABLED")).To(Succeed())
+				})
+
+				it("should not require anything at launch", func() {
+					require, err := reload.ShouldEnableLiveReload()
+					Expect(require).To(Equal(false))
+					Expect(err).NotTo(HaveOccurred())
+				})
+			})
+		})
+
+		context("failure cases", func() {
+			context("when BP_LIVE_RELOAD_ENABLED is not a valid boolean value", func() {
+				it.Before(func() {
+					Expect(os.Setenv("BP_LIVE_RELOAD_ENABLED", "hi")).To(Succeed())
+				})
+
+				it.After(func() {
+					Expect(os.Unsetenv("BP_LIVE_RELOAD_ENABLED")).To(Succeed())
+				})
+
+				it("should return an error", func() {
+					_, err := reload.ShouldEnableLiveReload()
+					Expect(err.Error()).To(Equal(`failed to parse BP_LIVE_RELOAD_ENABLED value hi: strconv.ParseBool: parsing "hi": invalid syntax`))
+				})
+			})
+		})
+	})
+
+	context("TransformReloadableProcesses", func() {
+		var (
+			originalProcess       packit.Process
+			reloadableProcessSpec reload.ReloadableProcessSpec
+
+			expectecdNonReloadableProcess packit.Process
+			expectedReloadableProcess     packit.Process
+		)
+
+		it.Before(func() {
+			reloadableProcessSpec = reload.ReloadableProcessSpec{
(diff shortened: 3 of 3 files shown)


robdimsdale:
On balance I don't think this needs to be added to packit, but I don't feel particularly strongly. If we do decide to add this, there are some changes I'd like to make which I address inline below.

> Is it worth hiding the implementation to reduce boilerplate?

I'm not convinced it is. I think the code in its current implemention is more confusing and not worth the encapsulation of re-use.

If there is consensus that we would like to go down this path, I would prefer to split up the responsibilities of checking if we should enable live reload from the actual buildplan requirement. Practically, this would look like a function called `shouldReload` and a public `var WatchExecRequirement`. I think this is cleaner than a function that has to be invoked multiple times with the `WatchExecRequirement` ignore.

> Should we use a fakeable func instead of a global func? This could help reduce testing boilerplate in consuming buildpacks.

Looking at the example implementation, the tests look essentially identical before and after this change, so this change wouldn't make the tests more tightly coupled to real behavior.

That being said, I already don't like that the unit tests in the consuming buildpack cannot be run in parallel due to relying on the state of the environment variable, so if we went down this path I would prefer to see a decoupling via interfaces and a fakeable function.

> Should this logic really exist in the watchexec buidpack? Could make some sense for that buildpack to expose some sort of SDK so that other buildpacks could know how to consume it.

This is interesting. I have been thinking about a related idea for other buildpacks that share common behavior but where that behavior doesn't belong in packit (e.g. common functionality across all of the python buildpacks). In the case of this particular functionality, the logic could live inside the `watchexec` buildpack, but I think there is a more general pattern we could explore of creating libraries that provide buildpack-specific (or buildpack family-specific) behavior. I haven't fully fleshed-out this idea yet, though.

joshuatcasey:
I'll close in a minute once I've done what you suggested, @robdimsdale . The purpose of this PR and a few others to Packit that I've made recently is to ease the buildpack authoring experience. Buildpacks that set start commands necessarily need to know whether process reloading is desired and set the start commands accordingly, but it feels like the implementation path is largely copy-and-paste.

> I would prefer to split up the responsibilities of checking if we should enable live reload from the actual buildplan requirement

This is a better idea. 

> I think there is a more general pattern we could explore of creating libraries that provide buildpack-specific

This is worth discussing. I feel like there are three places to put things right now: 
1. In the buildpack itself
2. In base packit
3. In the packit "extensions" (sbom, scribe, etc)

Feels like there could be something in between buildpacks and packit that could ease the authoring experience, since I do feel buildpacks do need to know about each others internals at least a little bit.

joshuatcasey:
@robdimsdale I'm curious what you think about the changes now. I've also updated the example consumer (paketo-buildpacks/go-build#310). 

I think this could actually make some sense in the `watchexec` buildpack, especially the new `TransformReloadableProcesses` func that transforms the start process into reloadable and non-reloadable start processes. Personally I don't like that various buildpacks (`go-build`, `poetry-run`, etc) have to be aware of what feels like watchexec internals to craft a start command.

reload/init_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	"github.com/onsi/gomega/format"
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+)
+
+func TestSBOM(t *testing.T) {
```
robdimsdale: I assume this should be: `TestReload` or similar?

reload/reload.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Shell will translate into --shell
+	Shell string
+
+	// Verbose will translate into -v
+	// If Verbose is 0, no -v arg will be provided.
+	// If Verbose is greater than 0, the appropriate number of v's will be added
+	// E.g. Verbose = 2 => -vv
+	// E.g. Verbose = 4 => -vvvv
+	Verbose int
```
robdimsdale: I'd probably prefer to call this: `VerbosityLevel` - as `Verbose` makes me think of booleans, rather than integers.

reload/reload.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	for _, watchDir := range spec.WatchDirs {
+		args = append(args, "--watch", watchDir)
+	}
+
+	args = append(args, "--shell", spec.Shell)
+
+	verbose := ""
+	for i := 0; i < spec.Verbose; i++ {
```
robdimsdale: I think we can refactor this slightly

```
	if verbosityLevel > 0 {
		args = append(args, "-"+strings.Repeat("v", verbosityLevel))
	}
```

reload/reload_test.go
```diff
@@ -0,0 +1,308 @@
...
+				Command: "my-command",
+				Args: []string{
+					"original-arg0",
+					"original-arg1",
+				},
+				Direct:           false,
+				Default:          false,
+				WorkingDirectory: "my-working-directory",
+			}
+			watchExecProcessSpec = reload.WatchExecProcessSpec{
```
robdimsdale: super small nit - but I would prefer to see the processes grouped together - and therefore i would move the watchexec spec to the top or bottom of the group of processes.

reload/reload_test.go
```diff
@@ -0,0 +1,308 @@
...
+			Expect(reloadableProcess).To(Equal(expectedReloadableProcess))
+		})
+
+		context("when direct is false and default is false", func() {
+			it.Before(func() {
+				originalProcess.Direct = false
+				originalProcess.Default = false
+			})
+
+			it("should return false, false, false, false", func() {
```
robdimsdale: nit: I think this test description could be clearer: "it should set direct and default to false on the returned processes"

robdimsdale: Same for the various tests below

robdimsdale: I also think for the rest of the test descriptions you can omit the 'direct/default is false' and just say: "when direct is true" / "when default is true" / "when direct and default are true".

I recognize this is more subjective, but i think it makes it easier to read:

- when x and y are false
- when x is true
- when y is true
- when x and y are true

I think this is ok because we don't have ternary states for the booleans - if it isn't `true` it must be `false`

reload/reload_test.go
```diff
@@ -0,0 +1,308 @@
...
+				nonReloadableProcess, reloadableProcess := reload.TransformReloadableProcesses(originalProcess, watchExecProcessSpec)
+
+				Expect(nonReloadableProcess).To(Equal(expectecNonReloadableProcess))
+				Expect(reloadableProcess).To(Equal(expectedReloadableProcess))
+			})
+		})
+	})
+
+	context("WatchExecRequirement", func() {
+		it("should require watchexec", func() {
```
robdimsdale: nit: i think this test would be clearer if it read: "it should require watchexec at launch"

robdimsdale (commented):
I like this. I think it encapsulates the logic for watchexec in a single place which will make it easier to roll out across more buildpacks.

I have a couple of refactors and renames (inline), but apart from that my only hangup is naming. I can't quite point to what I don't like about the naming, but once we commit to the names of public variables/functions we can't change them.

Give me a couple of days to let the names percolate and I'll see if I can identify what is giving me pause. If I can't articulate it (and hence suggest alternatives) I'd be happy to incorporate this feature.

robdimsdale:
Also, as far as whether these functions should be mockable or not - I'm happy to start with "not". We can always add an interface around concrete behavior later if we find it frustrating to test this in the buildpacks.

reload/init_test.go
joshuatcasey: You are correct.

reload/reload.go
joshuatcasey: I like that. I'm not really sure that any buildpacks actually use this, although I have done it once for debugging (https://github.com/paketo-buildpacks/go-build/pull/311#issuecomment-1125042614).

reload/reload.go
joshuatcasey: this is so much better

reload/reload_test.go
joshuatcasey: I'm not sure what you mean - move the test `context`?

reload/reload_test.go
joshuatcasey: you are correct.

joshuatcasey:
> I have a couple of refactors and renames (inline), but apart from that my only hangup is naming. I can't quite point to what I don't like about the naming, but once we commit to the names of public variables/functions we can't change them.

Part of the naming issue is that the names are a bit mixed between abstract (`reload`) and concrete (`watchexec`). The idea behind `reload` was to keep `watchexec` as an implementation detail, but it's pretty tightly coupled with `watchexec` and so I started to use that name.

> Also, as far as whether these functions should be mockable or not - I'm happy to start with "not". We can always add an interface around concrete behavior later if we find it frustrating to test this in the buildpacks.

There's a sense in which using these methods only eases half of the pain for the buildpack author - the author can use the funcs provided to generate the watchexec processes but must still know enough about watchexec to craft valid tests. So I see three possibilities:
1. Buildpack authors continue to write exhaustive watchexec tests as though the packit code were inline in the buildpack
2. Buildpack authors write a sample watchexec test just to verify that the buildpack is calling packit and so "doing the right thing"
3. Buildpack authors don't want to think about how watchexec works so they use the packit `reload` package, see what happens, and then hard-code the results as tests.

I think approach 2 is fine with me, since it implies that buildpack authors have done some of the analysis of how their process interacts with `watchexec`.

reload/reload_test.go
joshuatcasey: I'm open to this. Let me know what you think of the new names (commit coming in soon).

reload/reload_test.go
robdimsdale: no, sorry, just reorder the variable definitions so it looks like:

```
watchExecProcessSpec = ...
originalProcess = ...
expectedNonReloadableProcess = ...
expectedReloadableProcess = ...
```

joshuatcasey: oh I see. Done

joshuatcasey:
I'll mark this PR `Ready for review` although I'm tempted to think it should actually be in https://github.com/paketo-buildpacks/watchexec.

reload/reload_test.go
```diff
@@ -0,0 +1,333 @@
...
+				Args: []string{
+					"original-arg0",
+					"original-arg1",
+				},
+				Direct:           false,
+				Default:          false,
+				WorkingDirectory: "my-working-directory",
+			}
+
+			expectecNonReloadableProcess = packit.Process{
```
robdimsdale: typo: `expectecNonReloadableProcess` -> `expectedNonReloadableProcess`

reload/reload_test.go
robdimsdale: Exactly what I had in mind - thank you!

robdimsdale:
@joshuatcasey 

> Part of the naming issue is that the names are a bit mixed between abstract (`reload`) and concrete (`watchexec`). The idea behind `reload` was to keep `watchexec` as an implementation detail, but it's pretty tightly coupled with `watchexec` and so I started to use that name.

Yes, I think that is exactly what was making me uneasy. I agree that the implementation (watchexec) is tightly-coupled to the abstraction (live-reloading).

I think I would suggest renaming the `ShouldRequireWatchexec` to `LiveReloadEnabled`. This function has no references to watchexec, and equally, is the only place that references the environment variable `LIVE_RELOAD_ENABLED`.

I still don't feel great about `TransformReloadableProcesses` because this method is tightly coupled to the `watchexec` process implementation. However I can't identify a better name. I think it would have to be something like `TransformToWatchexecWrappedProcess` and that sounds really clunky.

For completeness - I think the naming of the struct `WatchExecProcessSpec` is fine. It is accurate and descriptive.

I was thinking about splitting the `reload` package into `reload` and `watchexec` to better separate out the interface and implementation, but I think that's too much boilerplate for little gain. However, if we went down the path of moving `watchexec` logic into the `watchexec` buildpack I would want to revisit this decision. I think it makes sense to have a live-reload abstraction in `packit` and a concrete implementation of this in `watchexec`. See below.

> I'll mark this PR Ready for review although I'm tempted to think it should actually be in https://github.com/paketo-buildpacks/watchexec.

On that note, I do like the idea of encapsulating this implementation somewhere more specific to `watchexec` - either the buildpack itself or a shared library. However, I would like to enumerate two or three situations where this pattern could be applied before making the change so we can study this more holistically.

> There's a sense in which using these methods only eases half of the pain for the buildpack author - the author can use the funcs provided to generate the watchexec processes but must still know enough about watchexec to craft valid tests. So I see three possibilities:
> 
> 1. Buildpack authors continue to write exhaustive watchexec tests as though the packit code were inline in the buildpack
> 2. Buildpack authors write a sample watchexec test just to verify that the buildpack is calling packit and so "doing the right thing"
> 3. Buildpack authors don't want to think about how watchexec works so they use the packit `reload` package, see what happens, and then hard-code the results as tests.
> 
> I think approach 2 is fine with me, since it implies that buildpack authors have done some of the analysis of how their process interacts with `watchexec`.

In practice I think it'll be some combination of 2 and 3 😂

sophiewigmore:
@joshuatcasey @robdimsdale I'm chiming in here late but I like this idea especially after looking at the code simplification in go-build. My 2 cents are that even though it's an uncommon pattern, I do think this code would fit more naturally as a part of the watchexec buildpack SDK

fg-j:
> @robdimsdale : I think it makes sense to have a live-reload abstraction in packit and a concrete implementation of this in watchexec. See below.

I agree. Given [active conversation](https://paketobuildpacks.slack.com/archives/C011S6EL49L/p1652334846460449) about the impact on buildpack authors/users of coupling `packit/sbom` to Syft's SBOM implementation, I'm wary of pushing logic into packit that's coupled to a specific live-reload implementation.

I think a live reload interface + separate implementation enhances buildpack authors' ability to customize buildpacks. They can use the implementation that comes with a buildpack or fork it and swap in their own. It enables buildpack authors to swap between live reload implementations based on user config at build time. 

There's some precedent for SDKS that multiple buildpacks use -- [libjvm](https://github.com/paketo-buildpacks/libjvm) is precisely this for the Paketo buildpacks that install different JVMs. Perhaps @dmikusa-pivotal can comment on the benefits/drawbacks of that library.

dmikusa:
>There's some precedent for SDKS that multiple buildpacks use -- [libjvm](https://github.com/paketo-buildpacks/libjvm) is precisely this for the Paketo buildpacks that install different JVMs. Perhaps @dmikusa-pivotal can comment on the benefits/drawbacks of that library.

tl;dr - I'm not sure this is a reason not to do this, but we're in the process of getting rid of libjvm. It has some challenges for us, so we're working to move away from it.

-------

Details:

The primary challenge is the dependency tree that it creates and the slowness to push out changes throughout that dependency tree. 

For us, libcnb is the base library, libpak builds on that and we have libjvm and libbs that build on libpak. Then tons of buildpacks that depend on all of those. A change to libcnb or libpak has to make its way up through the various layers, which is a slow process because we depend on dependabot to bump the other libraries & buildpacks and it has at best a 24h turnaround time to do that (or a lot of manual toil).

Beyond that...

- The interface exposed through libjvm covers about 90% of what's necessary, the other 10% has been hacked around with local modifications to the buildpack. It would be difficult to modify the API to cover 100% of what's necessary (breaking changes, more complicated API, lot of work, etc...), but at the same time the current hacks create pain and problems. 
- We still have 10 individual buildpacks, many of which do nothing more than initializing and delegating to the buildpack. There's still time spent maintaining those buildpacks, even though the code has been centralized in libjvm.

So it hasn't necessarily been a bad idea, it just hasn't panned out quite as we'd hoped and that paired with some other reasons means we're going to move away from it.

-------------------

Other thoughts about sharing code ...

- I worry a bit about one buildpack directly importing packages from another buildpack. I don't know of any Paketo buildpacks doing this at the moment. I worry about hidden assumptions there, i.e. I'm writing a buildpack not a library, have I made assumptions that might cause issues because of this. I also worry that it could import possibly more code than is necessary, depending on how the packages are structured for the buildpack.

- Right now, all of the utilities buildpacks are language agnostic. I think that would be something to discuss across the utilities team if we were to change that pattern.

- Utilities buildpacks like watchexec/upx/syft are really only intended to handle provides/requires & install a dependency. The consumption & usage of that dependency is up to the consuming buildpack. If we were to change that pattern, it would probably also be something to discuss across the utilities team as that changes what is required to understand to maintain those buildpacks.

- We're still dealing with a legacy of two golang libraries for writing buildpacks. This creates challenges when trying to write sharable/reusable code. We can't import both libraries so what you can put into shared code, especially if it were stored in a buildpack like Watchexec, would effectively need to be library agnostic.

robdimsdale:
@fg-j good point about adding more implementations to `packit` when we're starting to get feedback about the pain of that.

If we choose not to put this in packit, I would vote for creating a small single-purpose library rather than in the `watchexec` buildpack directly. We can always move code from that library into a buildpack if we decide that pattern makes sense. I'm wary about treating a buildpack as a library - it feels like a significant shift in responsibilities for a buildpack (for all the reasons @dmikusa-pivotal stated above) - whereas a small library has a well-established model of consumption and maintenance.

A separate library would also address the concerns about requiring the watchexec buildpack to depend on packit. A separate library could initially be used only by packit-based buildpacks, and then over time as packit's top-level interface and libcnb converge, the dependency becomes the common top-level interface rather than packit. At that point the watchexec buildpack can choose to consume this library without requiring a packit import.

reload/reload_test.go
joshuatcasey: thanks, fixed

joshuatcasey:
I appreciate all of the feedback so far. I think the best next step is to translate the code in this PR to a new repo, and to ensure there's a more generic `reload` interface with a `watchexec` implementation. I'll write this up as a RFC which I will propose soon and link here.

joshuatcasey:
See paketo-buildpacks/rfcs#203 for the continuing discussion. I'll close this PR since conversation should happen over there.

PR #363, opened 2022-07-26
Add debug logging for layer flags, build configuration

fg-j:
## Summary

This PR adds two new methods to the `scribe.Emitter`: `LayerFlags()` and `BuildConfiguration()`.

### LayerFlags
It'd be helpful for buildpack authors to log whether a layer is marked for build, cache, or launch in the debug output. With this PR, authors can add
```go
layer.Get("layer-name")
layer.Launch = true
layer.Cache = false
layer.Build = true

logger.LayerFlags(layer)
```
and when debug logging is enabled, the buildpack will log:
```
  Setting up layer 'layer-name'
    Available at app launch: true
    Available to other buildpacks:true
    Cached for rebuilds: false
```
This will allow buildpack users to understand more clearly why layer contents is/isn't available at a given stage.

❓  **I'm looking for feedback on the wording of these logs. Does this language accurately capture the behaviour of the flags without being too technical?**

### BuildConfiguration
I've also added a function that will enumerate the build configuration for a given buildpack. This is helpful for users trying to establish whether a desired configuration reached the buildpack.
Buildpack authors can add
```go
logger.BuildConfiguration(map[string]string{
     "BP_SOME_CONFIGURATION" : "value",
     "BP_OTHER_CONFIGURATION_ENABLED": "false"
})
```
and when debug logging is enabled, the buildpack will log:
```bash
  Build configuration:
    BP_OTHER_CONFIGURATION_ENABLED -> "false"
    BP_SOME_CONFIGURATION          -> "value"
```

This functionality handshakes well with the https://github.com/Netflix/go-env package, which we have been using in some buildpacks to extract environment variables into a struct. For example, in a `build.go`
```go
type Configuration struct {
  LogLevel string `env:"BP_LOG_LEVEL"`
  OtherConfig string `env:"BP_OTHER_CONFIG"`
}

func Build(config Configuration, logger scribe.Emitter) packit.BuildFunc{
    envSet, err := env.Marshal(&config) // unpack the config into an EnvSet
    logger.BuildConfiguration(envSet)
  }
}
```
will log:
```
  Build configuration:
    BP_LOG_LEVEL    -> "DEBUG"
    BP_OTHER_CONFIG -> "value"
```
For more on `EnvSet` and `env.Marshal()`, see [Netflix/go-env reference docs](https://pkg.go.dev/github.com/Netflix/go-env?utm_source=godoc).

⚠️  Note: To reuse the `NewFormattedMapFromEnvironment()`, I generalized the implementation to accept any `map[string]string` and added a case for map entries that don't have suffixes like `.override` or `.default`.

## Checklist

### DIFF
--- a/scribe/formatted_map.go
+++ b/scribe/formatted_map.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"sort"
 	"strings"
-
-	"github.com/paketo-buildpacks/packit/v2"
 )
 
 // A FormattedMap is a wrapper for map[string]interface{} to extend functionality.
@@ -46,10 +44,15 @@ func (m FormattedMap) String() string {
 
 // NewFormattedMapFromEnvironment take an environment and returns a
 // FormattedMap with the appropriate environment variable information added.
-func NewFormattedMapFromEnvironment(environment packit.Environment) FormattedMap {
+func NewFormattedMapFromEnvironment(environment map[string]string) FormattedMap {
 	envMap := FormattedMap{}
 	for key, value := range environment {
 		parts := strings.SplitN(key, ".", 2)
+
+		if len(parts) < 2 {
+			envMap[key] = value
+			continue
+		}
 
 		switch {
 		case parts[1] == "override" || parts[1] == "default":
--- a/scribe/emitter.go
+++ b/scribe/emitter.go
@@ -231,4 +231,5 @@ func (e Emitter) BuildConfiguration(envVars map[string]string) {
 	formatted := NewFormattedMapFromEnvironment(envVars)
 	e.Debug.Process("Build configuration:")
 	e.Debug.Subprocess(formatted.String())
+	e.Debug.Break()
 }
--- a/scribe/emitter.go
+++ b/scribe/emitter.go
@@ -162,7 +162,7 @@ func (e Emitter) LaunchProcesses(processes []packit.Process, processEnvs ...map[
 	e.Break()
 }
 
-// EnvironmentVariables take a layer and prints out a formatted table of the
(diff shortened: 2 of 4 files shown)


scribe/formatted_map.go
```diff
@@ -46,11 +44,16 @@ func (m FormattedMap) String() string {
 
 // NewFormattedMapFromEnvironment take an environment and returns a
 // FormattedMap with the appropriate environment variable information added.
-func NewFormattedMapFromEnvironment(environment packit.Environment) FormattedMap {
+func NewFormattedMapFromEnvironment(environment map[string]string) FormattedMap {
```
robdimsdale, line 47: I see that `packit.Environment` is a type alias for `map[string]string` so this change shouldn't be a breaking change for consumers. But I am curious as to why we want to remove this?

robdimsdale (approved):

ForestEckhardt (approved):

scribe/formatted_map.go
fg-j: See my note above in the PR body. This is so that we can pass either a `packit.Environment` or other string maps (like [the one that gets passed in to `emitter.BuildConfiguration()`](https://github.com/paketo-buildpacks/packit/blob/3a0f37af77a02ea5d2343412614e3a5cfce30f35/scribe/emitter.go#L230)). I figured this would be cleaner than converting the general string map argument of `BuildConfiguration()` into a `packit.Environment` before passing it into the constructor.

robdimsdale: gotcha - makes sense. I saw the comment in the PR body but I didn't connect it to the explanation in your comment, so thanks for that!

PR #365, opened 2022-07-29
Validate env var names before writing env var files

fg-j:
## Summary

In the process of using Netflix/go-env to log build configuration, I uncovered [this bug](https://github.com/Netflix/go-env/issues/23). I'm hoping to fix it upstream, but have to go through some bureaucracy to do it. For now, I've added validation to the `FormattedMap` constructor that should help out with this problem. 

This has also made me wonder whether we should have some validation in the `packit.Environment()` functions, since right now a user can easily add invalid env var values to the environment. I'm not sure how the lifecycle handles that situation today.

❓ @paketo-buildpacks/tooling-maintainers , do you think we should add validation to the `packit.Environment` methods?

## Checklist


ryanmoran:
Are we worried about it being less clear what is happening here since this code now just silently ignores bad inputs?

fg-j:
@ryanmoran You're right. I think it's confusing. I'll get around the bug in `go-env` a different way. But I do think some validation of environment variables could be useful.

I just played around with using `packit.Environment` to set environment variables with names that don't comply with the regex validation (e.g. setting `DOTNET ROOT` instead of `DOTNET_ROOT` or `RUNTIME=VERSION` instead of `RUNTIME_VERSION`. These get loaded into the environment without error, but evaluating them results in unexpected behaviour. For example, in a launch-time container,
```bash
cnb@5f362f11abd6:/workspace$ env
...
...
DOTNET_=ROOT=/workspace/.dotnet_root
...
...
cnb@5f362f11abd6:/workspace$ echo $DOTNET_=ROOT
ROOT=/workspace/.dotnet_root=ROOT
```

For that reason, I think the validation would make more sense as part of the [`EnvironmentWriter()`](https://github.com/paketo-buildpacks/packit/blob/43dbd221b26c79ce98e8d4b2e57e1aec798350c7/internal/environment_writer.go#L14). If that sounds reasonable, I'll shift the validation in this PR.

internal/environment_writer.go
```diff
@@ -22,6 +25,14 @@ func (w EnvironmentWriter) Write(dir string, env map[string]string) error {
 	}
 
 	for key, value := range env {
+		// this regex checks that map keys contain valid env var name characters,
+		// per https://pubs.opengroup.org/onlinepubs/9699919799/
+		validEnvVarRegex := regexp.MustCompile(`^[a-zA-Z_]{1,}[a-zA-Z0-9_]*$`)
```
ryanmoran: Not a huge performance concern, but since the regexp doesn't change over the loop, you can pull it out.

```suggestion
	// this regex checks that map keys contain valid env var name characters,
	// per https://pubs.opengroup.org/onlinepubs/9699919799/
	validEnvVarRegex := regexp.MustCompile(`^[a-zA-Z_]{1,}[a-zA-Z0-9_]*$`)
	for key, value := range env {
```

ryanmoran (approved):

ForestEckhardt (approved):

PR #372, opened 2022-08-03
Makes SPDX SBOM Reproducible

### LINKED ISSUE #368: Implement RFC 0049: Generate Reproducible SPDX SBOMs
## Describe the Enhancement

[RFC 0049](https://github.com/paketo-buildpacks/rfcs/blob/ded3562aefc2deb3b846761b8070f7be58bc68c8/text/0049-reproducible-spdx.md) proposes that Paketo buildpacks change how they generate SPDX SBOMs so that they are reproducible. This means modifying the `documentNamespace` and `created` fields in the SPDX JSON so their values are reproducible.

## Possible Solution

packit's `sbom` package should be modified so that SPDX SBOMs generated using the package:
- [ ] have a fixed timestamp in the `created` field. From the [RFC Implementation section](https://github.com/paketo-buildpacks/rfcs/blob/ded3562aefc2deb3b846761b8070f7be58bc68c8/text/0049-reproducible-spdx.md#implementation):
  > The value of `created` should be replaced with a default timestamp; if `$SOURCE_DATE_EPOCH` is set in the build environment, that time value should be used. If not, the zero-value time in Golang should be used.
- [ ] have a `documentNamespace` includes a UUID but is reproducible.  From the [RFC Implementation section](https://github.com/paketo-buildpacks/rfcs/blob/ded3562aefc2deb3b846761b8070f7be58bc68c8/text/0049-reproducible-spdx.md#implementation):
  > To produce a value for documentNamespace that (somewhat) uniquely identifies a given SBOM document, we should generate a [Version 5 SHA1 UUID](https://go-recipes.dev/how-to-generate-uuids-with-go-be3988e771a6) based on the hash of the struct stored in the FormattedReader's sbom field. We should replace the Syft-generated UUID with our reproducible one.
  - See the [example](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0049-reproducible-spdx.md#implementation:~:text=Overall%20the%20value%20of%20documentNamespace%20should%20be%20transformed%20from%3A) in the Implementation section for further details.

The RFC suggests modifying [`sbom.FormattedReader.Read()`](https://github.com/paketo-buildpacks/packit/blob/429f8e4370b9579e1c3340ede29a82f58152136d/sbom/formatted_reader.go#L41) to edit the SPDX JSON after it's generated. See [this spike of packit](https://github.com/paketo-buildpacks/packit/compare/reproducible-spdx) for an example of hashing an SBOM to create a UUID and rewriting JSON output with the result.

## Motivation

If SBOMs are not reproducible, they make buildpack-built images non-reproducible. Buildpack users and buildpack authors using packit shouldn't compromise build reproducibility just to get image SBOMs.

ForestEckhardt:
- Writes custom documentNamespace and use SOURCE_DATE_EPOCH for created
  timestamp when available

Resolves #368 

## Checklist

### DIFF
--- a/sbom/formatted_reader.go
+++ b/sbom/formatted_reader.go
@@ -131,7 +131,7 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
 						if err != nil {
 							return 0, fmt.Errorf("failed to parse SOURCE_DATE_EPOCH: %w", err)
 						}
-						creationInfo["created"] = time.Unix(sde, 0)
+						creationInfo["created"] = time.Unix(sde, 0).UTC()
 					}
 					spdxOutput[k] = creationInfo
 				}
--- a/sbom/formatted_reader.go
+++ b/sbom/formatted_reader.go
@@ -93,14 +93,14 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
 				return 0, fmt.Errorf("failed to modify SPDX SBOM for reproducibility: %w", err)
 			}
 
-			// Makes the creationInfo reproducible to a hash can be taken for the
+			// Makes the creationInfo reproducible so a hash can be taken for the
 			// documentNamespace
 			if creationInfo, ok := spdxOutput["creationInfo"].(map[string]interface{}); ok {
 				creationInfo["created"] = time.Time{} // This is the zero-valued time
 
-				source_date_epoch := os.Getenv("SOURCE_DATE_EPOCH")
-				if source_date_epoch != "" {
-					sde, err := strconv.ParseInt(source_date_epoch, 10, 64)
+				sourceDateEpoch := os.Getenv("SOURCE_DATE_EPOCH")
+				if sourceDateEpoch != "" {
+					sde, err := strconv.ParseInt(sourceDateEpoch, 10, 64)
 					if err != nil {
 						return 0, fmt.Errorf("failed to parse SOURCE_DATE_EPOCH: %w", err)
 					}
--- a/sbom/formatted_reader.go
+++ b/sbom/formatted_reader.go
@@ -2,7 +2,6 @@ package sbom
 
 import (
 	"bytes"
-	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -17,7 +16,6 @@ import (
 	"github.com/anchore/syft/syft"
 	"github.com/anchore/syft/syft/sbom"
 	"github.com/google/uuid"
-	"github.com/mitchellh/hashstructure/v2"
 )
 
 // FormattedReader outputs the SBoM in a specified format.
@@ -88,37 +86,15 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
 		// Makes SPDX SBOM more reproducible, see
 		// https://github.com/paketo-buildpacks/packit/issues/368 for more details.
 		if f.format.ID() == "spdx-2-json" {
-			hash, err := hashstructure.Hash(f.sbom.syft, hashstructure.FormatV2, nil)
-			if err != nil {
-				// not tested
-				return 0, fmt.Errorf("failed to modify SPDX SBOM for reproducibility: %w", err)
-			}
-			hashBytes := make([]byte, binary.MaxVarintLen64)
-			binary.PutUvarint(hashBytes, hash)
-
 			var spdxOutput map[string]interface{}
 
 			err = json.Unmarshal(output, &spdxOutput)
 			if err != nil {
 				return 0, fmt.Errorf("failed to modify SPDX SBOM for reproducibility: %w", err)
 			}
-			if namespace, ok := spdxOutput["documentNamespace"].(string); ok {
-				uri, err := url.Parse(namespace)
-				if err != nil {
-					// not tested
-					return 0, err
-				}
 
-				uri.Host = "paketo.io"
-				uri.Path = strings.Replace(uri.Path, "syft", "packit", 1)
-				oldBase := filepath.Base(uri.Path)
-				source, _, _ := strings.Cut(oldBase, "-")
-				newBase := fmt.Sprintf("%s-%s", source, uuid.NewSHA1(uuid.NameSpaceURL, hashBytes))
-				uri.Path = strings.Replace(uri.Path, oldBase, newBase, 1)
-
-				spdxOutput["documentNamespace"] = uri.String()
-			}
-
+			// Makes the creationInfo reproducible to a hash can be taken for the
+			// documentNamespace
 			if creationInfo, ok := spdxOutput["creationInfo"].(map[string]interface{}); ok {
 				creationInfo["created"] = time.Time{} // This is the zero-valued time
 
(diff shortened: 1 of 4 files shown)


sbom/formatted_reader_test.go
```diff
@@ -94,24 +96,77 @@ func testFormattedReader(t *testing.T, context spec.G, it spec.S) {
...
+			it.After(func() {
+				Expect(os.Setenv("SOURCE_DATE_EPOCH", original)).To(Succeed())
+			})
+
+			it("produces an SBOM with the given timestamp", func() {
+				buffer := bytes.NewBuffer(nil)
+				_, err := io.Copy(buffer, sbom.NewFormattedReader(bom, sbom.SPDXFormat))
+				Expect(err).NotTo(HaveOccurred())
+
+				var spdxOutput spdxOutput
```
fg-j: The unit tests for the FormattedReader need to be run sequentially because we're setting the env var, which is global state.

sbom/formatted_reader_test.go
```diff
@@ -94,24 +96,77 @@ func testFormattedReader(t *testing.T, context spec.G, it spec.S) {
...
+				Expect(spdxOutput.Packages[0].Name).To(Equal("collapse-white-space"), buffer.String())
+				Expect(spdxOutput.Packages[1].Name).To(Equal("end-of-stream"), buffer.String())
+				Expect(spdxOutput.Packages[2].Name).To(Equal("insert-css"), buffer.String())
+				Expect(spdxOutput.Packages[3].Name).To(Equal("once"), buffer.String())
+				Expect(spdxOutput.Packages[4].Name).To(Equal("pump"), buffer.String())
+				Expect(spdxOutput.Packages[5].Name).To(Equal("wrappy"), buffer.String())
+
+				// Ensure documentNamespace and creationInfo.created have reproducible values
+				Expect(spdxOutput.DocumentNamespace).To(Equal("https://paketo.io/packit/dir/testdata-37a75df4-7cb2-5384-820a-d9ac961acccb"), buffer.String())
+				Expect(spdxOutput.CreationInfo.Created).To(Equal(time.Unix(1659551872, 0).UTC()), buffer.String())
```
fg-j: I'm seeing this test fail on my local:
```
        Expected
            <time.Time>: 2022-08-03T14:37:52-04:00
        to equal
            <time.Time>: 2022-08-03T18:37:52Z
```

sbom/formatted_reader.go
```diff
@@ -78,6 +87,62 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
...
+				if k == "creationInfo" {
+					creationInfo := spdxOutput["creationInfo"].(map[string]interface{})
+					source_date_epoch := os.Getenv("SOURCE_DATE_EPOCH")
+					if source_date_epoch == "" {
+						creationInfo["created"] = time.Time{} // This is the zero-valued time
+					} else {
+						sde, err := strconv.ParseInt(source_date_epoch, 10, 64)
+						if err != nil {
+							// not tested
+							return 0, fmt.Errorf("failed to parse SOURCE_DATE_EPOCH: %w", err)
```
fg-j: Can we unit test this case by setting a weird value for `SOURCE_DATE_EPOCH`?

sbom/formatted_reader.go
```diff
@@ -78,6 +87,62 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
...
+					uri.Path = strings.Replace(uri.Path, oldBase, newBase, 1)
+
+					spdxOutput[k] = uri.String()
+				}
+				if k == "creationInfo" {
+					creationInfo := spdxOutput["creationInfo"].(map[string]interface{})
+					source_date_epoch := os.Getenv("SOURCE_DATE_EPOCH")
+					if source_date_epoch == "" {
+						creationInfo["created"] = time.Time{} // This is the zero-valued time
+					} else {
```
fg-j:
```suggestion
                                         creationInfo["created"] = time.Time{} // This is the zero-valued time

					source_date_epoch := os.Getenv("SOURCE_DATE_EPOCH")
					if source_date_epoch != "" {
```

fg-j: A small simplification of the logic.

fg-j (changes requested):

sbom/formatted_reader.go
```diff
@@ -78,6 +87,62 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
...
+                                         creationInfo["created"] = time.Time{} // This is the zero-valued time
+
+					source_date_epoch := os.Getenv("SOURCE_DATE_EPOCH")
+					if source_date_epoch != "" {
+						sde, err := strconv.ParseInt(source_date_epoch, 10, 64)
+						if err != nil {
+							// not tested
+							return 0, fmt.Errorf("failed to parse SOURCE_DATE_EPOCH: %w", err)
+						}
+						creationInfo["created"] = time.Unix(sde, 0)
```
fg-j:
```suggestion
						creationInfo["created"] = time.Unix(sde, 0).UTC()
```
Resolves https://github.com/paketo-buildpacks/packit/pull/372#discussion_r937045461

sbom/formatted_reader_test.go
ForestEckhardt: Tests that set env vars have been made sequential

sbom/formatted_reader.go
ForestEckhardt: A failure case unit test has been added

fg-j:
@sophiewigmore could you take a peek at this when you have a chance?

fg-j (dismissed):

ForestEckhardt:
Bump @paketo-buildpacks/tooling-maintainers

sbom/formatted_reader.go
```diff
@@ -78,6 +87,61 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
...
+			hashBytes := make([]byte, binary.MaxVarintLen64)
+			binary.PutUvarint(hashBytes, hash)
+
+			var spdxOutput map[string]interface{}
+
+			err = json.Unmarshal(output, &spdxOutput)
+			if err != nil {
+				return 0, fmt.Errorf("failed to modify SPDX SBOM for reproducibility: %w", err)
+			}
+			for k := range spdxOutput {
```
ryanmoran: Is there a reason to loop over the keys here rather than just doing something like the following?

```go
if namespace, ok := output["documentNamespace"].(string); ok {
  // do work
}
```

sbom/formatted_reader.go
```diff
@@ -78,6 +87,61 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
...
+		// Makes SPDX SBOM more reproducible, see
+		// https://github.com/paketo-buildpacks/packit/issues/368 for more details.
+		if f.format.ID() == "spdx-2-json" {
+			hash, err := hashstructure.Hash(f.sbom.syft, hashstructure.FormatV2, nil)
+			if err != nil {
+				// not tested
+				return 0, fmt.Errorf("failed to modify SPDX SBOM for reproducibility: %w", err)
+			}
+			hashBytes := make([]byte, binary.MaxVarintLen64)
+			binary.PutUvarint(hashBytes, hash)
```
ryanmoran: If you wanted remove the need to pull in this extra third-party library, you could just use stdlib. Maybe something like the following:

```go
var spdxOutput map[string]interface{}
err = json.Unmarshal(output, &spdxOutput)
if err != nil {
	return 0, fmt.Errorf("failed to modify SPDX SBOM for reproducibility: %w", err)
}

delete(spdxOutput, "documentNamespace")
delete(spdxOutput, "creationInfo")

hash := sha1.New()
err = json.NewEncoder(hash).Encode(spdxOutput)
if err != nil {
	return 0, fmt.Errorf("failed to checksum SPDX document: %w", err)
}

checksumSha1 := hash.Sum(nil)
```

ryanmoran (commented):
I'm generally fine with this changeset. I've added a couple thoughts on some changes you might consider.

sbom/formatted_reader.go
ForestEckhardt: I personally am not too concerned about pulling in `mitchellh/hashstructure`.

fg-j: I ended up going with the other approach in my spike because it felt weird to create a SHA that doesn't really correspond to anything -- it's not the SHA of the final SBOM, nor is it the SHA of the original SBOM or the package list in the SBOM. But I do like that this approach avoids pulling in a third-party lib.
 @ForestEckhardt do you have thoughts on this?

fg-j: FWIW `mitchellh/hashstructure` doesn't have external dependencies, we're not adding much more bloat to `packit`.

ForestEckhardt: I have just pushed an implementation that does not use  `mitchellh/hashstructure` that I think might be more straight forward. Please take a look and let me known what you think!

fg-j (dismissed):

fg-j (approved):

PR #379, opened 2022-08-09
Marshal CycloneDX/SPDX SBOM with indentation

sophiewigmore:
## Summary

The [latest packit release ](https://github.com/paketo-buildpacks/packit/releases/tag/v2.4.0) performs an extra modification step for the CycloneDX/SPDX SBOMs to make them reproducible. This change causes the output SBOMs to format without any indentation, all on one line. 

This PR changes the `json.Marshal` to `json.MarshalIndent` for purely cosmetic reasons, so that the output SBOMs match other SBOMs with spacing. 

The SBOM will come out looking like:
```
{
        "bomFormat": "CycloneDX",
        "components": [
                {
                        "cpe": "cpe:2.3:a:haxx:curl:7.84.0:*:*:*:*:*:*:*"
...
```
Instead of:
```
{"bomFormat":"CycloneDX","components":[{"cpe":"cpe:2.3:a:haxx:curl:7.84.0:*:*:*:*:*:*:*", ...
```

## Checklist

### DIFF
--- a/sbom/formatted_reader.go
+++ b/sbom/formatted_reader.go
@@ -77,7 +77,9 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
 
 			delete(cycloneDXOutput, "serialNumber")
 
-			output, err = json.MarshalIndent(cycloneDXOutput, "", "\t")
+			// Indent with a two spaces, as they do in CycloneDX:
+			// https://github.com/CycloneDX/cyclonedx-go/blob/429d353cfcdbfedf367f597cbdde2a840ebf29df/encode.go#L44
+			output, err = json.MarshalIndent(cycloneDXOutput, "", "  ")
 			if err != nil {
 				return 0, fmt.Errorf("failed to modify CycloneDX SBOM for reproducibility: %w", err)
 			}
@@ -133,7 +135,9 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
 				spdxOutput["documentNamespace"] = uri.String()
 			}
 
-			output, err = json.MarshalIndent(spdxOutput, "", "\t")
+			// Indent with a single space, as they do in SPDX:
+			// https://github.com/anchore/syft/blob/1344889766743beb736aafdfb29266910b738fbb/internal/formats/spdx22json/encoder.go#L16
+			output, err = json.MarshalIndent(spdxOutput, "", " ")
 			if err != nil {
 				return 0, fmt.Errorf("failed to modify SPDX SBOM for reproducibility: %w", err)
 			}
--- a/sbom/formatted_reader.go
+++ b/sbom/formatted_reader.go
@@ -77,7 +77,7 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
 
 			delete(cycloneDXOutput, "serialNumber")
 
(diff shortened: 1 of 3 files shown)


fg-j:
Can we expand the tests in some way to assert that SBOMs come out pretty-printed?

sophiewigmore:
@fg-j if you happen to know a nicer way of doing this, let me know. I'd like to avoid having to check the entirety of the JSON content which is what you run into with `MatchJSON`.

sophiewigmore:
I also cannot figure out what is wrong with the linter..

ForestEckhardt:
@sophiewigmore We are using the deprecated `CPE` field from `postal` in `sbom`

sophiewigmore:
Gah!! Different linter versions are behaving differently. It's working locally

sophiewigmore:
@ForestEckhardt I think `Error: Process completed with exit code 137.` is a Docker issue not a linter issue, related to RAM

sbom/formatted_reader.go
```diff
@@ -133,7 +133,7 @@ func (f *FormattedReader) Read(b []byte) (int, error) {
 				spdxOutput["documentNamespace"] = uri.String()
 			}
 
-			output, err = json.Marshal(spdxOutput)
+			output, err = json.MarshalIndent(spdxOutput, "", "\t")
```
fg-j: nit:
I see that in syft's SBOM formatting implementation, for SPDX json, they [use](https://github.com/anchore/syft/blob/1344889766743beb736aafdfb29266910b738fbb/internal/formats/spdx22json/encoder.go#L16) 
```go
	enc := json.NewEncoder(output)
	// prevent > and < from being escaped in the payload
	enc.SetEscapeHTML(false)
	enc.SetIndent("", " ")
```
and for CycloneDX JSON, they [use the cyclonedx-go package](https://github.com/anchore/syft/blob/1344889766743beb736aafdfb29266910b738fbb/internal/formats/cyclonedxjson/encoder.go#L13), which [uses](https://github.com/CycloneDX/cyclonedx-go/blob/429d353cfcdbfedf367f597cbdde2a840ebf29df/encode.go#L44)
```go
	if j.pretty {
		encoder.SetIndent("", "  ")
	}
```


So, with this implementation, we're effectively replacing spaces in the SBOM files with tabs. For consistency, I think it makes sense to use spaces instead. It's a bit weird that one JSON format uses single-space indentation and the other uses two. I figure we just pick one and stick with it.

sophiewigmore: @fg-j Good idea - can you clarify - you want us to use spaces instead but just pick either one space or two for all formats? What would stop us from just using one space for SPDX and two for CycloneDx so we match their underlying representations?

sophiewigmore: I have opted for using one space for SPDX and two for Cyclonedx so that our output SBOMs just match the upstream as much as possible.

fg-j: Fair point. We may as well match the existing implementations since we're already calling `json.MarshalIndent` separately for SPDX and CycloneDX.

sophiewigmore:
The linter has been fixed in #380, I'll rebase this PR against main after that PR gets merged in

fg-j (approved):

PR #389, opened 2022-09-08
Adds checksum and source-checksum field in dependencies

ForestEckhardt:
Given the move a federated buildpack dependencies it seems like it might be prudent for us to allow for different checksum algorithms to be used so that we can better adapt to checksums given by upstream providers. I have put together a quick POC to gauge the interest as well as prove out the possibility of the approach. I would love feedback on this!

### DIFF
--- a/postal/service_test.go
+++ b/postal/service_test.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"crypto/sha256"
+	"crypto/sha512"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -355,6 +356,7 @@ version = "this is super not semver"
 	context("Deliver", func() {
 		var (
 			dependencySHA string
+			hash512       string
 			layerPath     string
 			deliver       func() error
 		)
@@ -394,6 +396,9 @@ version = "this is super not semver"
 
 			sum := sha256.Sum256(buffer.Bytes())
 			dependencySHA = hex.EncodeToString(sum[:])
+
+			sum512 := sha512.Sum512(buffer.Bytes())
+			hash512 = hex.EncodeToString(sum512[:])
 
 			transport.DropCall.Returns.ReadCloser = io.NopCloser(buffer)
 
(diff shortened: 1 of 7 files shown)


postal/service_test.go
```diff
@@ -441,6 +446,53 @@ version = "this is super not semver"
...
+				deliver = func() error {
+					return service.Deliver(
+						postal.Dependency{
+							ID:     "some-entry",
+							Stacks: []string{"some-stack"},
+							URI:    "some-entry.tgz",
+							Checksum: postal.DependencyChecksum{
+								Algorithm: "SHA512",
+								Hash:      hash512,
+							},
```
ryanmoran: You could maybe avoid the extra TOML structure if you encoded the algorithm as a prefix like `sha512:<hash>` into the checksum string.

So, instead of
```toml
[metadata.dependencies.checksum]
  algorithm = "SHA512"
  hash = "some-sha512-hash"
```

you'd have
```toml
checksum = "sha512:some-sha512-hash"
```

robdimsdale: +1 - as an example, this is how docker defines the digest of an image. It seems more idiomatic as well as being less structure to handle.

ForestEckhardt: Hmmm that is interesting it means that I would keep the `ValidatedReader.New()` function signature the same. Let me try it out.

joshuatcasey: I like the idea and the direction. Is the eventual goal to remove `sha256` as a standalone field? We should allow the new `checksum` field to store `sha256`, and I wonder if the validator could check that `sha256` and `checksum` are equal to each other in the case that both are present, and then eventually we can remove the `sha256` field. It's similar to doing a database migration moving one column to another column in two stages, so that no single change removes data that isn't already in another field.

ForestEckhardt: I just pushed up an approach using the formatted checksum string. The only drawback is that `cargo.ValidatedReader.New()` now returns an error. This is because I cannot set up the `hash.Hash` struct in Read because it get called multiple times in any given `Read()` call meaning that the `hash.Hash` struct would get remade on each call making them empty every time. If someone has a better approach I would love to hear it.

ryanmoran: A couple options:
1) Create a new signature that is expected to be used to construct ValidatedReaders with these types of checksums.
2) Defer the error until you Read using the ValidatedReader. You could implement an `UnsupportedHash` type that conforms to the `hash.Hash` interface and returns an error message saying that type is unsupported.

ForestEckhardt: I tried doing option number 2 but I was running into a problem where the `hash.Hash` was empty. It look like `ValidatedReader.Read()` gets called twice and if I initialize the `hash.Hash` in the `Read()` function it gets reset.

ForestEckhardt: I am not sure that I understand the logic behind option 1. Is the idea that the current `cargo.NewValidatedReader()` should just continue to only support sha256 by default and then we have another new that is different but supports a wider range of algorithms? If that is the case I am not sure I understand why the current function should exist if it covers the old use case.

ryanmoran: Your concern is about maintaining backwards compatibility with the method signature. You always have the choice of making a new method with a different signature. I'm just calling out that that is an option.

ForestEckhardt: I am not sure I care to much about maintaining backwards compatibility entirely, I was just pointing it out as a draw back. I did however I did get the `errorHash` concept working and looking pretty good so I will push that up here in a second and see if it makes things hard to grock.

joshuatcasey:
Could we do the same thing for `SourceSHA256`?

ryanmoran:
Should we mark the `SHA256` and `SourceSHA256` fields as deprecated?

ForestEckhardt:
@ryanmoran I can try it out and see what happens.

ryanmoran (approved):

PR #396, opened 2022-09-16
Support usage of `checksum` field in legacy sbom

sophiewigmore:
## Summary

Add support for the new `checksum` field in the legacy Paketo SBOM. 
If the value of the checksum is malformed (i.e. not in  format `algorithm`:`hash`), then the `algorithm` field will be set to `UNKNOWN`, and the `hash` value will be an empty string.

## Checklist

### DIFF
--- a/paketosbom/sbom.go
+++ b/paketosbom/sbom.go
@@ -60,7 +60,7 @@ func GetBOMChecksumAlgorithm(alg string) (algorithm, error) {
 		}
 	}
 
-	return "", fmt.Errorf("failed to get supported BOM checksum algorithm: %s is not valid", alg)
+	return UNKNOWN, fmt.Errorf("failed to get supported BOM checksum algorithm: %s is not valid", alg)
 }
 
 const (
@@ -76,4 +76,5 @@ const (
 	BLAKE2B512 algorithm = "BLAKE2b-512"
 	BLAKE3     algorithm = "BLAKE3"
 	MD5        algorithm = "MD5"
+	UNKNOWN    algorithm = "UNKNOWN"
 )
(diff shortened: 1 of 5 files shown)


sophiewigmore:
@ryanmoran there was no direct usage of the `SHA256` field in the newer SBOM `GenerateFromDependency` implementation, but I did update the tests to use the `Checksum` field (which had no effect on the outcome of the tests)

paketosbom/sbom.go
```diff
@@ -76,4 +76,5 @@ const (
 	BLAKE2B512 algorithm = "BLAKE2b-512"
 	BLAKE3     algorithm = "BLAKE3"
 	MD5        algorithm = "MD5"
+	UNKNOWN    algorithm = "UNKNOWN"
```
sophiewigmore, line 79: @paketo-buildpacks/tooling-maintainers I thought the addition of the `UNKNOWN` algorithm would make it clearer in the output when something has gone wrong since the `GenerateBillOfMaterials` function does not return an error of any kind. I can remove this if you think it's unnecessary though.

postal/service.go
```diff
@@ -240,3 +255,22 @@ func (s Service) GenerateBillOfMaterials(dependencies ...Dependency) []packit.BO
...
+
+	// A well-formed checksum field (algorithm:hash) takes precedence over a SHA256 field
+	checksumParts := strings.Split(checksumField, ":")
+	if len(checksumParts) > 1 {
+		algorithm = checksumParts[0]
+		hash = checksumParts[1]
+	}
+
+	return algorithm, hash
+}
```
ryanmoran:
```suggestion
func determineChecksum(checksumField, sha256Field string) (string, string) {
	// A well-formed checksum field (algorithm:hash) takes precedence over a SHA256 field
	algorithm, hash, found := strings.Cut(checksumField, ":")
	if found {
		return algorithm, hash
	}

	if len(sha256Field) > 0 {
		return "SHA256", sha256Field
	}

	return "", ""
}
```

ryanmoran (approved):

PR #399, opened 2022-09-22
Refactors out common Checksum concept

ryanmoran:
## Summary

We have some common code for splitting and comparing checksums across `cargo` and `postal`. This factors those pieces out into a common place.

## Checklist


sophiewigmore:
I can't comment on it directly for some reason, but there is a linting error in postal/service.go to remove function `determineChecksum`

sophiewigmore:
Also, this seems like a reasonable change but I wanted to point out that introducing the new Checksum type will be breaking for `jam`, since it uses the `cargo.ConfigMetadataDependency` internally

cargo/checksum_test.go
```diff
@@ -0,0 +1,54 @@
...
+		type testCaseType struct {
+			c1     string
+			c2     string
+			result bool
+		}
+
+		for _, tc := range []testCaseType{
+			{"", "", true},
+			{"c", "c", true},
+			{"c", "c", true},
```
fg-j: Is this the same test case twice or am I misunderstanding the test setup?

cargo/checksum_test.go
```diff
@@ -0,0 +1,54 @@
...
+			{"sha256:c", "c", true},
+			{"c", "sha256:c", true},
+			{"md5:c", "md5:c", true},
+			{"md5:c", ":c", false},
+			{":", ":", true},
+			{":c", ":c", true},
+			{"", "c", false},
+			{"c", "", false},
+			{"c", "z", false},
+			{"md5:c", "sha256:c", false},
```
fg-j, line 35: For completeness, perhaps it also makes sense to check when the alg is specified the same and the hash is different:
```suggestion
			{"md5:c", "sha256:c", false},
			{"md5:c", "md5:d", false},
```

cargo/validated_reader.go
```diff
@@ -26,28 +25,20 @@ type errorHash struct {
...
-
-	switch splitChecksum[0] {
+	switch checksum.Algorithm() {
 	case "sha256":
 		hash = sha256.New()
 	case "sha512":
 		hash = sha512.New()
 	default:
-		return ValidatedReader{hash: errorHash{err: fmt.Errorf("unsupported algorithm %q: the following algorithms are support [sha256, sha512]", splitChecksum[0])}}
+		return ValidatedReader{hash: errorHash{err: fmt.Errorf("unsupported algorithm %q: the following algorithms are support [sha256, sha512]", checksum.Algorithm())}}
```
fg-j:
```suggestion
		return ValidatedReader{hash: errorHash{err: fmt.Errorf("unsupported algorithm %q: the following algorithms are supported [sha256, sha512]", checksum.Algorithm())}}
```

fg-j (commented):
A couple nitpicks/clarifications. But overall, LGTM!

ryanmoran:
> I can't comment on it directly for some reason, but there is a linting error in postal/service.go to remove function `determineChecksum`

Good catch. I meant to remove this.

cargo/checksum_test.go
ryanmoran: Woops. I should have read the existing tests more closely when I refactored this.

ryanmoran:
> Also, this seems like a reasonable change but I wanted to point out that introducing the new Checksum type will be breaking for `jam`, since it uses the `cargo.ConfigMetadataDependency` internally

Ok. I rolled back the changes that might have been breaking. We still use `Checksum` internally and users can cast their checksum string to a `Checksum` to use its helper methods, so I still think this results in a cleaner implementation.

sophiewigmore (approved):

PR #402, opened 2022-09-30
Deprioritize wildcard stacks when resolving dependency.

robdimsdale:
## Summary

This PR fixes a bug in dependency resolution where sometimes a dependency with a wildcard stack (i.e. `*` stack) would be selected over a dependency with a more specific, matching, stack (e.g. `io.buildpacks.stacks.jammy`). This PR also returns an error when multiple dependencies with the same version support wildcard stacks, as this should not be supported.

There is nothing in the paketo RFCs about this behavior that I've found. I think it was just assumed to work this way based on various conversations I've had with contributors/consumers.

## Use Cases

In the case where there are multiple dependencies for a given version, and one of the dependencies contains a wildcard stack, the dependency with wildcard stack support should be prioritized below the one with the more specific stack. This allows for situations like providing pre-compiled dependencies for a specific stack but with the ability to fall back to compiling from source for any stack that is not not explicitly specified.

In the case where multiple dependencies with the same version each support wildcard stacks, we should fail and return an error. This is because there is no way to conceptually choose between these dependencies, and as such this is a logical error. To expand on this, ask the question: how would a human choose between two dependencies with the same version (e.g. `1.2.3`) and the same stack selector (i.e. `*`). Arguably, we could expand the error handling to include multiple dependencies with the same version and stack information for non-wildcard stacks, but I think that is less likely to happen.

### Aside

For the case where there are multiple dependencies with wildcard stacks, it would technically be possible to select a dependency if there was one dependency with support for both a specific, _matching_ stack and a wildcard stack (e.g. [`io.buildpacks.stacks.matching`, `*`]) and another dependency with support for both a _non-matching_ stack and a wildcard stack (e.g. `[io.buildpacks.stacks.non-matching`, `*`]), or just a wildcard stack (e.g. `*`).

Similarly, one could expand that situation to include a dependency with a set of stacks where one is matching, plus the wildcard (e.g. `io.buildpacks.stacks.matching`, `io.buildpacks.stacks.non-matching`, `*`) and a dependency where one contains a set of stacks that do not match (e.g. `io.buildpacks.stacks.non-matching`, `io.buildpacks.stacks.also-non-matching`, `*`).

However, these are all edge cases that I do not think we need to handle just yet. I'm not fully convinced they even make conceptual sense - what would it mean to have two dependencies that can support any stack (via the wildcard) but not the same specific stack? What should the prioritization be here?

## Checklist

### DIFF
--- a/postal/service.go
+++ b/postal/service.go
@@ -132,13 +132,82 @@ func (s Service) Resolve(path, id, version, stack string) (Dependency, error) {
 		)
 	}
 
+	stacksForVersion := map[string][]string{}
+
+	for _, dep := range compatibleVersions {
+		stacksForVersion[dep.Version] = append(stacksForVersion[dep.Version], dep.Stacks...)
+	}
+
+	for version, stacks := range stacksForVersion {
+		count := stringSliceElementCount(stacks, "*")
+		if count > 1 {
+			return Dependency{}, fmt.Errorf("multiple dependencies support wildcard stack for version: %q", version)
+		}
+	}
+
 	sort.Slice(compatibleVersions, func(i, j int) bool {
-		iVersion := semver.MustParse(compatibleVersions[i].Version)
-		jVersion := semver.MustParse(compatibleVersions[j].Version)
-		return iVersion.GreaterThan(jVersion)
+		iDep := compatibleVersions[i]
+		jDep := compatibleVersions[j]
+
+		jVersion := semver.MustParse(jDep.Version)
+		iVersion := semver.MustParse(iDep.Version)
+
+		if !iVersion.Equal(jVersion) {
+			return iVersion.GreaterThan(jVersion)
+		}
+
+		iStacks := iDep.Stacks
+		jStacks := jDep.Stacks
+
+		// If either dependency supports the wildcard stack, it has lower
+		// priority than a dependency that only supports a more specific stack.
+		// This is true regardless of whether or not the dependency with
+		// wildcard stack support also supports other stacks
+		//
+		// If is an error to have multiple dependencies with the same version
+		// and wildcard stack support.
+		// This is tested for above, and we would not enter this sort function
+		// in this case
+
+		if stringSliceContains(iStacks, "*") {
+			return false
+		}
+
+		if stringSliceContains(jStacks, "*") {
+			return true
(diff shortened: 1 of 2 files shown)


ryanmoran (approved):
This looks great. I'm pretty sure that when this code was originally written `*` stack IDs did not exist and so we'd never contemplated their impact on how we might choose the right dependency.

At the same time, I believe that nearly all of the cases you outline are really examples of kludges over a tragically anemic API for defining stack compatibility. In almost every case where `*` stack is specified, we almost never mean literally any stack. We just mean those stacks without specific IDs that we can know ahead of time.

I also think that in some instances where we have a dependency for a specific stack and another for `*` with the same version number, we may actually mean different things. In the primary case we have now, we're really saying one of these dependencies is precompiled for this stack, the other requires a compilation process. They represent conceptually different things. Its not that the `*` stack dependency can just be dropped onto the filesystem and run. In that case, comparing them seems odd. Perhaps they should really be defined by different dependency IDs so that the buildpack isn't comparing them at all?

robdimsdale:
@ryanmoran sure, I agree with everything you lay out. In principle I think we should explore the idea of different dependency IDs for pre-compiled/source dependencies.

However, the current behavior is causing issues in the Python ecosystem, because sometimes packit is resolving the "compile-from-source" dependency when it should be picking a pre-compiled dependency. The effects of this include: test timeouts because builds take much longer, test failures because python is compiled with different flags, etc.

You could argue that we shouldn't have released the python buildpack with `*` stack support without thinking through these ramifications, but we made an assumption that packit would resolve the dependencies in the same order, and now we'd have to revert that feature to fix the impacted buildpacks.

In my ideal case, we would merge this PR to fix the currently-broken buildpacks, and we can then take our time to explore the ramifications of using different dependency IDs.

I think it's reasonable for packit to support the behavior proposed in this PR (i.e. deprioritizing wildcard stacks) even if we then decide in the future to deprecate support for that feature.

Thoughts?

ryanmoran:
Sorry. I should have been clearer in my feedback. I am not opposed to merging this. Of course we should do the expedient thing. I'm just outlining some other thoughts about what we are really dealing with. I'll go ahead and merge this now.

PR #422, opened 2022-11-03
Enable support for checksum in postal dependency mappings

### LINKED ISSUE #420: FindDependencyMapping can not find any dependency mapping if only new checksum attribute is used
The Dependency attribute `SHA256` is deprecated and `checksum` should be used instead.

The `Deliver` method checks for an existing dependency mapping only with the `SHA256` attribute. If it is not present, because only the new `checksum` is used, no dependency mapping can be found.

Possible solution:
Use the SHA256 from the `checksum` to find dependency mappings. For backward compatibility use `SHA256` attribute if `checksum` is not present.

```
func (s Service) Deliver(dependency Dependency, cnbPath, layerPath, platformPath string) error {
	dependencyMappingURI, err := s.mappingResolver.FindDependencyMapping(dependency.SHA256, platformPath)
	if err != nil {
		return fmt.Errorf("failure checking for dependency mappings: %s", err)
	}

	if dependencyMappingURI != "" {
		dependency.URI = dependencyMappingURI
	}
```

Is my observation correct or am I missing something?

sophiewigmore:
## Summary

Resolves #420 

Introduces support for the `checksum` field for postal dependency mappings.
The new behaviour is backwards compatible to work with the `sha256` field as well as the new `checksum` field we now support in the `buildpack.toml`.

`FindDependencyMapping` now has slightly different behaviour:
  - If a binding is given in the form of just a `hash` (no algorithm), it will assume it is of algorithm type `sha256` in order to be backwards compatible with how users expect to provide a `dependency-mapping` binding. **Binding lookup will fail if the dependency checksum from the `buildpack.toml` is not of type `sha256`.**
  - We now allow users to provide a binding in the form of `algorithm:hash` (like the checksum field we support), so that bindings can still be provided for dependencies that don't use the `sha256` algorithm.

### Open Question for @paketo-buildpacks/tooling-maintainers 
Do we need to make modifications to https://github.com/paketo-buildpacks/rfcs/blob/main/text/0010-dependency-mappings.md, since we now support non-sha256 checksums as the binding value?

## Checklist

### DIFF
--- a/postal/internal/dependency_mappings.go
+++ b/postal/internal/dependency_mappings.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/paketo-buildpacks/packit/v2/cargo"
 	"github.com/paketo-buildpacks/packit/v2/servicebindings"
 )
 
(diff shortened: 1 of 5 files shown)


postal/internal/dependency_mappings.go
```diff
@@ -23,19 +24,31 @@ func NewDependencyMappingResolver(bindingResolver BindingResolver) DependencyMap
...
+// If the binding is given in the form of `algorithm:hash`, compare it to the full `checksum` input
+func (d DependencyMappingResolver) FindDependencyMapping(checksum, platformDir string) (string, error) {
 	bindings, err := d.bindingResolver.Resolve("dependency-mapping", "", platformDir)
 	if err != nil {
 		return "", fmt.Errorf("failed to resolve 'dependency-mapping' binding: %w", err)
 	}
 
 	for _, binding := range bindings {
-		if uri, ok := binding.Entries[sha256]; ok {
+		hash := cargo.Checksum(checksum).Hash()
```
ryanmoran: You can move this outside the loop so it doesn't get re-evaluated on each iteration.

sophiewigmore:
RFC addendum: https://github.com/paketo-buildpacks/rfcs/blob/0e13fd55abae75a46c4ee73a2ba46988bc7b0379/text/0000-dependency-mappings-addendum.md

ryanmoran (approved):

fg-j (approved):

PR #425, opened 2022-11-10
Return a typed error when no dependencies are found during resolution.

### LINKED ISSUE #415: Enable consumers to skip failure to resolve dependencies
## Describe the Enhancement

I would like to find a better solution for callers of `postal.Service.Resolve()` to identify that no dependencies were resolved without failing the build.

Currently this is possible, but it relies on inspecting the returned error text. That is brittle, and I'd like to find a better solution.

## Motivation

The specific use-case I have in mind is splitting out precompiled dependencies from source dependencies (e.g. in the cpython buildpack) via different IDs/names and resolving them independently. In this scenario, we would like to be able to bypass the failure associated with resolving zero precompiled dependencies, while still failing if there are zero source dependencies resolved.

## Possible Solutions

Below are multiple ways to achieve this, ordered by my most-preferred to my least-preferred solution.

1. Return newly-created specific error types to `postal` (e.g. `ErrNoDependenciesResolved`, `ErrMultipleDependenciesResolved`) which would allow consumers to introspect the error as follows:
    ```
    precompiledDependency, err := service.Resolve()
    if err != nil && !errors.Is(err, postal.ErrNoDependenciesResolved) {
      // fail
    }
    ```
    **Pros:** 
    * Does not change the interface of the function (i.e. it is backwards-compatible)
    * There is precedent for this in golang (e.g. `if err != nil && !errors.Is(err, fs.ErrExist){ ... }`)
    
    **Cons:**
    * This is the most complex option for consumers
    * Adds the most code to the `postal` package
    * Introspecting on error types is an ugly way to manage control flow for non-failure cases
      
1. Add number of dependencies resolved as part of the result (e.g. `func Resolve() (Dependency, int, error))`)
     ```
    precompiledDependency, count, err := service.Resolve()
    if err != nil && count != 0 {
      // fail
    }
    ```
    **Pros:**
    * Fairly easy to read and understand
    
    **Cons:** 
    * Not backwards-compatible
    * Somewhat inelegant

1. Allow users to pass in some config to skip the failure entirely (e.g. `func Resolve([existing args] , skipNoDepsFail bool) (Dependency, error)` or `func Resolve([existing args], config SomeNewConfigType) (Dependency, error))`)
     ```
    precompiledDependency, err := service.Resolve([existing args], true)
    if err != nil { // err is nil when zero dependencies are resolved
      // fail
    }
    ```
    **Pros:**
    * Simple for consumers; keeps consumer flow clean
    * Fairly easy to read and understand
    
    **Cons:** 
    * Not backwards-compatible
    * Quite inelegant.

robdimsdale:
## Summary

This PR adds a typed error to be returned when no dependencies are found during dependency resolution. This is a backwards-compatible change that allows callers to introspect on the error type (via `errors.As()`) in order to handle this case separately from other failures. See #415 for more details on when this case arises.

Callers who do not care about handling this failure mode will continue to see the existing behavior - the error message is unchanged.

## Use Cases

Resolves #415 

## Design/Architecture discussion

There are two ways to achieve the goal of returning specific, typed, errors agreed upon in #415:
1. a sentinel error (similar to [`os.ErrNotExist`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/os/error.go;l=24))
2. a new instantiation of a custom error type.

I opted for the second option - returning an instantiation of a custom error type, despite the fact that the sentinel pattern is more familiar. I chose this because I wanted to be able to instantiate the error with specific data (like the discovered supported versions) in order to preserve the existing error message. The sentinel value pattern would force us to return the same error message in all cases, which I felt was less useful.

The main difference for callers is the first option would require using `errors.Is` and the second option would require using `errors.As`.

## Checklist

### DIFF
--- a/postal/service.go
+++ b/postal/service.go
@@ -37,6 +37,26 @@ type MappingResolver interface {
 	FindDependencyMapping(checksum, platformDir string) (string, error)
 }
 
+// ErrNoDeps is a typed error indicating that no dependencies were resolved during Service.Resolve()
+//
+// errors can be tested against this type with: errors.As()
+type ErrNoDeps struct {
+	id                string
+	version           string
+	stack             string
+	supportedVersions []string
+}
+
+// Error implements the error.Error interface
+func (e *ErrNoDeps) Error() string {
+	return fmt.Sprintf("failed to satisfy %q dependency version constraint %q: no compatible versions on %q stack. Supported versions are: [%s]",
+		e.id,
+		e.version,
+		e.stack,
+		strings.Join(e.supportedVersions, ", "),
+	)
+}
+
 // Service provides a mechanism for resolving and installing dependencies given
 // a Transport.
 type Service struct {
@@ -123,13 +143,7 @@ func (s Service) Resolve(path, id, version, stack string) (Dependency, error) {
 	}
 
 	if len(compatibleVersions) == 0 {
-		return Dependency{}, fmt.Errorf(
-			"failed to satisfy %q dependency version constraint %q: no compatible versions on %q stack. Supported versions are: [%s]",
-			id,
-			version,
-			stack,
-			strings.Join(supportedVersions, ", "),
-		)
+		return Dependency{}, &ErrNoDeps{id, version, stack, supportedVersions}
 	}
 
 	stacksForVersion := map[string][]string{}
--- a/postal/service_test.go
+++ b/postal/service_test.go
@@ -420,8 +420,10 @@ version = "1.2.3"
 			})
 
 			context("when the entry version constraint cannot be satisfied", func() {
-				it("returns an error with all the supported versions listed", func() {
+				it("returns a typed error with all the supported versions listed", func() {
+					expectedErr := &postal.ErrNoDeps{}
 					_, err := service.Resolve(path, "some-entry", "9.9.9", "some-stack")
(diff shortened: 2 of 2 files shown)


sophiewigmore (approved):

sophiewigmore:
looks good to me! I'm happy with the path you took. @paketo-buildpacks/tooling-maintainers any other feedback before we merge this in?

ryanmoran (approved):

robdimsdale:
Alright, looks like we got 👍 from half the tooling maintainers so I'm going to merge this now.

postal/service.go
```diff
@@ -37,6 +37,26 @@ type MappingResolver interface {
 	FindDependencyMapping(checksum, platformDir string) (string, error)
 }
 
+// ErrNoDeps is a typed error indicating that no dependencies were resolved during Service.Resolve()
+//
+// errors can be tested against this type with: errors.As()
+type ErrNoDeps struct {
```
robdimsdale, line 43: For posterity, I have recently learned that the idiom for these types of _structured errors_ is to use the `Error` suffix. So if I were writing this again, I'd name it something like `NoDepsError`.

The `Err` prefix is idiomatic for sentinel errors - package-level errors that do not get instantiated with specific data.

https://abhinav.github.io/presentation-go-patterns-2022/#/structured-errors/0

PR #433, opened 2022-12-07
Add stack extension support

### LINKED ISSUE #432: Buildpack API 0.9: Stack Extension Support
## Describe the Enhancement

Implement any support required for this [CNB RFC](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md) to enable users to make Stack Extensions using `packit`.

## Motivation

The APIs for Buildpacks and Stack Extensions are very similar. We can likely leverage the same codebase to do both.

c0d1ngm0nk3y:
fixes #432 

## Summary

This adds stack extension support according to the spec (https://github.com/buildpacks/spec/blob/main/image_extension.md)

## Use Cases

We used generic to have as much code reuse between `Detect` of a buildpack and `Detect` of an extension. If we would not use generics, this would mean to copy the logic which is basically the same. The main differences are:

* `CNB_BUILDPACK_DIR` vs. `CNB_EXTENSION_DIR`
* `buildpack.toml` vs. `extension.toml`

Note: The current state reuses the `BuildpackInfo` struct (renaming it), although `extension.toml` does not contain `SBOMFormats`.

We tried to not break any user with this change.

## Checklist

### DIFF
--- a/detect_test.go
+++ b/detect_test.go
@@ -18,8 +18,8 @@ import (
 
 func testDetect(t *testing.T, context spec.G, it spec.S) {
 	var (
-		Expect = NewWithT(t).Expect
-
+		Expect      = NewWithT(t).Expect
+		err         error
 		workingDir  string
 		tmpDir      string
 		platformDir string
@@ -33,7 +33,6 @@ func testDetect(t *testing.T, context spec.G, it spec.S) {
 	)
 
 	it.Before(func() {
-		var err error
 		workingDir, err = os.Getwd()
 		Expect(err).NotTo(HaveOccurred())
 
@@ -56,21 +55,6 @@ func testDetect(t *testing.T, context spec.G, it spec.S) {
 
 		binaryPath = filepath.Join(cnbDir, "bin", "detect")
 
-		bpTOMLContent := []byte(`
-api = "0.5"
-[buildpack]
-  id = "some-id"
-  name = "some-name"
-  version = "some-version"
-  clear-env = false
-`)
-		Expect(os.WriteFile(filepath.Join(cnbDir, "buildpack.toml"), bpTOMLContent, 0600)).To(Succeed())
-
-		planDir, err = os.MkdirTemp("", "buildplan.toml")
-		Expect(err).NotTo(HaveOccurred())
-
-		planPath = filepath.Join(planDir, "buildplan.toml")
-
 		exitHandler = &fakes.ExitHandler{}
 	})
 
@@ -78,61 +62,146 @@ api = "0.5"
 		Expect(os.Chdir(workingDir)).To(Succeed())
 		Expect(os.RemoveAll(tmpDir)).To(Succeed())
 		Expect(os.RemoveAll(cnbDir)).To(Succeed())
-		Expect(os.RemoveAll(planDir)).To(Succeed())
 		Expect(os.RemoveAll(platformDir)).To(Succeed())
 		Expect(os.Unsetenv("CNB_STACK_ID")).To(Succeed())
 	})
 
-	context("when providing the detect context to the given DetectFunc", func() {
-		it("succeeds", func() {
-			var context packit.DetectContext
+	context("when detect within an extension", func() {
 
-			packit.Detect(func(ctx packit.DetectContext) (packit.DetectResult, error) {
-				context = ctx
+		it.Before(func() {
+			extTOMLContent := []byte(`
+api = "0.9"
+[extension]
+  id = "some-id"
+  name = "some-name"
+  version = "some-version"
+`)
+			Expect(os.WriteFile(filepath.Join(cnbDir, "extension.toml"), extTOMLContent, 0600)).To(Succeed())
+			Expect(os.Setenv("CNB_EXTENSION_DIR", cnbDir))
+			planDir, err = os.MkdirTemp("", "buildplan.toml")
+			Expect(err).NotTo(HaveOccurred())
 
-				return packit.DetectResult{}, nil
-			}, packit.WithArgs([]string{binaryPath, platformDir, planPath}))
+			planPath = filepath.Join(planDir, "buildplan.toml")
 
-			Expect(context).To(Equal(packit.DetectContext{
-				WorkingDir: tmpDir,
-				CNBPath:    cnbDir,
-				Platform: packit.Platform{
-					Path: platformDir,
-				},
-				BuildpackInfo: packit.BuildpackInfo{
-					ID:      "some-id",
-					Name:    "some-name",
-					Version: "some-version",
-				},
-				Stack: stackID,
-			}))
+		})
+
+		it.After(func() {
+			Expect(os.RemoveAll(planDir)).To(Succeed())
+			Expect(os.Unsetenv("CNB_EXTENSION_DIR")).To(Succeed())
+		})
+
+		context("when providing the detect context to the given DetectFunc", func() {
+			it("succeeds", func() {
+				var context packit.DetectContext
+
+				packit.Detect(func(ctx packit.DetectContext) (packit.DetectResult, error) {
+					context = ctx
+
+					return packit.DetectResult{}, nil
+				}, packit.WithArgs([]string{binaryPath, platformDir, planPath}), packit.WithExitHandler(exitHandler))
+
+				Expect(exitHandler.ErrorCall.CallCount).To(Equal(0))
+				Expect(context).To(Equal(packit.DetectContext{
+					WorkingDir: tmpDir,
+					CNBPath:    cnbDir,
+					Platform: packit.Platform{
+						Path: platformDir,
+					},
+					BuildpackInfo: packit.BuildpackInfo{
+						ID:      "some-id",
+						Name:    "some-name",
+						Version: "some-version",
+					},
+					Info: packit.Info{
+						ID:      "some-id",
+						Name:    "some-name",
+						Version: "some-version",
+					},
+					Stack: stackID,
+				}))
+			})
 		})
 	})
 
-	it("writes out the buildplan.toml", func() {
-		packit.Detect(func(packit.DetectContext) (packit.DetectResult, error) {
-			return packit.DetectResult{
-				Plan: packit.BuildPlan{
-					Provides: []packit.BuildPlanProvision{
-						{Name: "some-provision"},
+	context("when detect within a buildpack", func() {
+		it.Before(func() {
+			bpTOMLContent := []byte(`
+api = "0.5"
+[buildpack]
+  id = "some-id"
+  name = "some-name"
+  version = "some-version"
+  clear-env = false
+`)
+			Expect(os.WriteFile(filepath.Join(cnbDir, "buildpack.toml"), bpTOMLContent, 0600)).To(Succeed())
+
+			planDir, err = os.MkdirTemp("", "buildplan.toml")
+			Expect(err).NotTo(HaveOccurred())
+
+			planPath = filepath.Join(planDir, "buildplan.toml")
+		})
+
+		it.After(func() {
+			Expect(os.RemoveAll(planDir)).To(Succeed())
+		})
+
+		context("when providing the detect context to the given DetectFunc", func() {
+			it("succeeds", func() {
+				var context packit.DetectContext
+
+				packit.Detect(func(ctx packit.DetectContext) (packit.DetectResult, error) {
+					context = ctx
+
+					return packit.DetectResult{}, nil
+				}, packit.WithArgs([]string{binaryPath, platformDir, planPath}), packit.WithExitHandler(exitHandler))
+
+				Expect(exitHandler.ErrorCall.CallCount).To(Equal(0))
+				Expect(context).To(Equal(packit.DetectContext{
+					WorkingDir: tmpDir,
+					CNBPath:    cnbDir,
+					Platform: packit.Platform{
+						Path: platformDir,
 					},
-					Requires: []packit.BuildPlanRequirement{
-						{
-							Name: "some-requirement",
-							Metadata: map[string]string{
-								"version":  "some-version",
-								"some-key": "some-value",
+					BuildpackInfo: packit.BuildpackInfo{
+						ID:      "some-id",
+						Name:    "some-name",
+						Version: "some-version",
+					},
+					Info: packit.Info{
+						ID:      "some-id",
+						Name:    "some-name",
+						Version: "some-version",
+					},
+					Stack: stackID,
+				}))
+			})
+		})
+
+		it("writes out the buildplan.toml", func() {
+			packit.Detect(func(packit.DetectContext) (packit.DetectResult, error) {
+				return packit.DetectResult{
+					Plan: packit.BuildPlan{
+						Provides: []packit.BuildPlanProvision{
+							{Name: "some-provision"},
+						},
+						Requires: []packit.BuildPlanRequirement{
+							{
+								Name: "some-requirement",
+								Metadata: map[string]string{
+									"version":  "some-version",
+									"some-key": "some-value",
+								},
 							},
 						},
 					},
-				},
-			}, nil
-		}, packit.WithArgs([]string{binaryPath, platformDir, planPath}))
+				}, nil
+			}, packit.WithArgs([]string{binaryPath, platformDir, planPath}), packit.WithExitHandler(exitHandler))
 
-		contents, err := os.ReadFile(planPath)
-		Expect(err).NotTo(HaveOccurred())
+			Expect(exitHandler.ErrorCall.CallCount).To(Equal(0))
+			contents, err := os.ReadFile(planPath)
+			Expect(err).NotTo(HaveOccurred())
 
-		Expect(string(contents)).To(MatchTOML(`
+			Expect(string(contents)).To(MatchTOML(`
 [[provides]]
   name = "some-provision"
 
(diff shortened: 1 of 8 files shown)


ryanmoran:
Sorry this hasn't had much feedback yet. I took the time today to read through the diff and so far this looks okay to me. I want to try it out to see how the API feels first. I won't have time to do that until next week though.

sophiewigmore:
+1

sophiewigmore:
will take a look at this first thing tomorrow

sophiewigmore:
The changes to `Detect` will be breaking for buildpacks in that detection as we have it now in many places will fail with: 
```
./detect.go:20:47: cannot use generic type packit.DetectFunc[R packit.DetectResultInterface] without instantiation
```
so we'll have to update that everywhere.
However, I think the way you've done this is smart and I think I'd prefer that over completely duplicating the code

sophiewigmore:
Should `BuildpackInfoLicense` inside of `Info` become `InfoLicense`?

detect_test.go
```diff
@@ -91,8 +91,9 @@ api = "0.5"
 				context = ctx
 
 				return packit.DetectResult{}, nil
-			}, packit.WithArgs([]string{binaryPath, platformDir, planPath}))
+			}, packit.WithArgs([]string{binaryPath, platformDir, planPath}), packit.WithExitHandler(exitHandler))
```
sophiewigmore: I feel like we should add a test for the extension.toml code path as well for completeness

generate.go
```diff
@@ -0,0 +1,161 @@
...
+	cnbPath := os.Getenv("CNB_EXTENSION_DIR")
+	outputPath := os.Getenv("CNB_OUTPUT_DIR")
+	platformPath := os.Getenv("CNB_PLATFORM_DIR")
+
+	var info struct {
+		APIVersion string `toml:"api"`
+		Info       Info   `toml:"extension"`
+	}
+
+	extensionYAML := filepath.Join(cnbPath, "extension.toml")
```
sophiewigmore: nit
```suggestion
	extensionTOML := filepath.Join(cnbPath, "extension.toml")
```

generate.go
```diff
@@ -0,0 +1,161 @@
...
+
+	var plan BuildpackPlan
+	_, err = toml.DecodeFile(planPath, &plan)
+	if err != nil {
+		config.exitHandler.Error(err)
+		return
+	}
+
+	cnbPath := os.Getenv("CNB_EXTENSION_DIR")
+	outputPath := os.Getenv("CNB_OUTPUT_DIR")
```
sophiewigmore, line 108: Maybe we should add a check here to ensure that the `outputPath` is definitely not set to the `<app>` directory

sophiewigmore:
Most of the comments so far are just nitpicks, I don't have any comments on the API design (yet)

generate.go
ryanmoran: What's the concern here? The `CNB_OUTPUT_DIR` env var comes from the lifecycle. I feel like we trust what the lifecycle passes to the extension.

generate.go
```diff
@@ -0,0 +1,161 @@
...
+// will be parsed and persisted by the Build function and returned to the
+// lifecycle at the end of the build phase execution.
+type GenerateResult struct {
+	// TODO: Config
+	ExtendConfig ExtendConfig
+
+	// TODO: BuildDockerfile
+	BuildDockerfile *Dockerfile
+	// TODO: RunDockerfile
+	RunDockerfile *Dockerfile
```
ryanmoran: Rather than using pointers to what ultimately are `string` values, what if we expected the user to pass an `io.Reader`. This way they can pass file descriptors or any other thing that can be wrapped in the `io.Reader` interface.

generate.go
```diff
@@ -0,0 +1,161 @@
...
+
+// TODO: BuildFunc is the definition of a callback that can be invoked when the Build
+// function is executed. Buildpack authors should implement a BuildFunc that
+// performs the specific build phase operations for a buildpack.
+type GenerateFunc func(GenerateContext) (GenerateResult, error)
+
+// TODO: BuildContext provides the contextual details that are made available by the
+// buildpack lifecycle during the build phase. This context is populated by the
+// Build function and passed to BuildFunc during execution.
+type GenerateContext struct {
```
ryanmoran, line 21: Rather than duplicating the input here, can we alias the type? At some point these may diverge, but we can handle breaking them apart at that point.

ryanmoran (commented):
I'm still mulling over whether or not we should expose a generics contract to end users. While I think about that, I've proposed some changes that might help to streamline a couple parts of the API.

generate.go
c0d1ngm0nk3y: > can we alias the type?

It is not identical, is it? The `BuildContext` contains the layers and the `GenerateContext` not and the `DetectContext` contains the plan which the `GenerateContext` doesn't.

c0d1ngm0nk3y:
> However, I think the way you've done this is smart and I think I'd prefer that over completely duplicating the code

Should we remove the generics again or accept the breaking change? And should we then move to `v3`? Strictly speaking, it is breaking, but the fix is just adding `[ExtensionDetectResult]` to the signature.

WDYT?

sophiewigmore:
@c0d1ngm0nk3y personally I think accepting the breaking change is worth it. I think the other tooling maintainers should weigh in though, which might not be for a bit as many of them are out now until the new year

ryanmoran:
I think that in the long run, I'd like to avoid using generics in the primary API. To achieve this, I think we need to accept a different, but probably less controversial change: we just use the same `DetectFunc` for both buildpacks and stack extensions. This would mean that extension authors could accidentally require dependencies from their extensions which is an invalid operation. I think that this would be an acceptable tradeoff to keep a much simpler API.

In the worst case, a stack extension author accidentally requires something and they see this message from the lifecycle when they try to use their extension:
```
===> DETECTING
[detector] Warning: Platform requested experimental feature 'Dockerfiles'
[detector] ======== Error: samples/tree@1.2.3 ========
[detector] extension samples/tree outputs "requires" which is not allowed
[detector] err:  samples/tree@1.2.3
[detector] ERROR: No buildpack groups passed detection.
[detector] ERROR: Please check that you are running against the correct path.
[detector] ERROR: failed to detect: no buildpacks participating
ERROR: failed to build: executing lifecycle. This may be the result of using an untrusted builder: failed with status code: 20
```

Also considering the API, I think we can live with a single `Run` method that takes either a `BuildFunc` or a `GenerateFunc` as its second argument. We can achieve this with an interface type.

Here's a diff of what these changes might look like:

```diff
diff --git a/build.go b/build.go
index 17f656c..8616915 100644
--- a/build.go
+++ b/build.go
@@ -21,6 +21,8 @@ import (
 // performs the specific build phase operations for a buildpack.
 type BuildFunc func(BuildContext) (BuildResult, error)
 
+func (bf BuildFunc) _BuildOrGenerate() {}
+
 // BuildContext provides the contextual details that are made available by the
 // buildpack lifecycle during the build phase. This context is populated by the
 // Build function and passed to BuildFunc during execution.
diff --git a/detect.go b/detect.go
index 4bce61f..966aff5 100644
--- a/detect.go
+++ b/detect.go
@@ -12,7 +12,7 @@ import (
 // DetectFunc is the definition of a callback that can be invoked when the
 // Detect function is executed. Buildpack authors should implement a DetectFunc
 // that performs the specific detect phase operations for a buildpack.
-type DetectFunc[R DetectResultInterface] func(DetectContext) (R, error)
+type DetectFunc func(DetectContext) (DetectResult, error)
 
 // DetectContext provides the contextual details that are made available by the
 // buildpack lifecycle during the detect phase. This context is populated by
@@ -55,29 +55,10 @@ type DetectResult struct {
 	Plan BuildPlan
 }
 
-func (r DetectResult) plan() interface{} {
-	return r.Plan
-}
-
-type ExtensionDetectResult struct {
-	Plan ExtensionBuildPlan
-}
-
-func (r ExtensionDetectResult) plan() interface{} {
-	return r.Plan
-}
-
-type DetectResultInterface interface {
-	plan() interface{}
-}
-
-var _ DetectResultInterface = DetectResult{}
-var _ DetectResultInterface = ExtensionDetectResult{}
-
 // Detect is an implementation of the detect phase according to the Cloud
 // Native Buildpacks specification. Calling this function with a DetectFunc
 // will perform the detect phase process.
-func Detect[R DetectResultInterface](f DetectFunc[R], options ...Option) {
+func Detect(f DetectFunc, options ...Option) {
 	config := OptionConfig{
 		exitHandler: internal.NewExitHandler(),
 		args:        os.Args,
@@ -154,7 +135,7 @@ func Detect[R DetectResultInterface](f DetectFunc[R], options ...Option) {
 	}
 	defer file.Close()
 
-	err = toml.NewEncoder(file).Encode(result.plan())
+	err = toml.NewEncoder(file).Encode(result.Plan)
 	if err != nil {
 		config.exitHandler.Error(err)
 		return
diff --git a/generate.go b/generate.go
index 38aaff4..03fe5f7 100644
--- a/generate.go
+++ b/generate.go
@@ -15,6 +15,8 @@ import (
 // performs the specific build phase operations for a buildpack.
 type GenerateFunc func(GenerateContext) (GenerateResult, error)
 
+func (gf GenerateFunc) _BuildOrGenerate() {}
+
 // TODO: BuildContext provides the contextual details that are made available by the
 // buildpack lifecycle during the build phase. This context is populated by the
 // Build function and passed to BuildFunc during execution.
diff --git a/run.go b/run.go
index c39d8f5..9e1d538 100644
--- a/run.go
+++ b/run.go
@@ -8,50 +8,41 @@ import (
 	"github.com/paketo-buildpacks/packit/v2/internal"
 )
 
-// Run combines the invocation of both build and detect into a single entry
-// point. Calling Run from an executable with a name matching "build" or
-// "detect" will result in the matching DetectFunc or BuildFunc being called.
-func Run(detect DetectFunc[DetectResult], build BuildFunc, options ...Option) {
-	run(map[string]func(){
-		"detect": func() {
-			Detect(detect, options...)
-		},
-		"build": func() {
-			Build(build, options...)
-		},
-	}, options...)
+type BuildOrGenerateFunc interface {
+	_BuildOrGenerate()
 }
 
-// TODO: Run combines the invocation of both build and detect into a single entry
+// Run combines the invocation of both build and detect into a single entry
 // point. Calling Run from an executable with a name matching "build" or
 // "detect" will result in the matching DetectFunc or BuildFunc being called.
-func RunExtension(detect DetectFunc[ExtensionDetectResult], generate GenerateFunc, options ...Option) {
-	run(map[string]func(){
-		"detect": func() {
-			Detect(detect, options...)
-		},
-		"generate": func() {
-			Generate(generate, options...)
-		},
-	}, options...)
-}
-
-func run(m map[string]func(), options ...Option) {
+func Run(detect DetectFunc, buildOrGenerate BuildOrGenerateFunc, options ...Option) {
 	config := OptionConfig{
 		exitHandler: internal.NewExitHandler(),
 		args:        os.Args,
 	}
 
-	for _, option := range options {
-		config = option(config)
+	var (
+		build    BuildFunc
+		generate GenerateFunc
+	)
+
+	switch f := buildOrGenerate.(type) {
+	case BuildFunc:
+		build = f
+	case GenerateFunc:
+		generate = f
 	}
 
 	phase := filepath.Base(config.args[0])
-
-	f, ok := m[phase]
-	if !ok {
+	switch phase {
+	case "detect":
+		Detect(detect, options...)
+	case "build":
+		Build(build, options...)
+	case "generate":
+		Generate(generate, options...)
+	default:
 		config.exitHandler.Error(fmt.Errorf("failed to run buildpack: unknown lifecycle phase %q", phase))
-		return
 	}
-	f()
+
 }
```

With both of these changes, we don't need to bump the major version since the API is not changing in a backwards-incompatible way. We also keep the API as small as seems reasonable.

Thoughts?

c0d1ngm0nk3y:
> I think that in the long run, I'd like to avoid using generics in the primary API.

What is the reason? Complexity?

> This would mean that extension authors could accidentally require dependencies from their extensions which is an invalid operation. I think that this would be an acceptable tradeoff to keep a much simpler API.

But we would loose the compiler to be able to detect the misuse of the API. Like you pointed out, the user would rather run into a runtime issue.
 
> In the worst case, a stack extension author accidentally requires something and they see this message from the lifecycle when they try to use their extension:
> 
> ```
> ===> DETECTING
> [detector] Warning: Platform requested experimental feature 'Dockerfiles'
> [detector] ======== Error: samples/tree@1.2.3 ========
> [detector] extension samples/tree outputs "requires" which is not allowed
> [detector] err:  samples/tree@1.2.3
> [detector] ERROR: No buildpack groups passed detection.
> [detector] ERROR: Please check that you are running against the correct path.
> [detector] ERROR: failed to detect: no buildpacks participating
> ERROR: failed to build: executing lifecycle. This may be the result of using an untrusted builder: failed with status code: 20
> ```
> 
> Also considering the API, I think we can live with a single `Run` method that takes either a `BuildFunc` or a `GenerateFunc` as its second argument. We can achieve this with an interface type.

Sounds good.

> Thoughts?

Do we really wanna keep the API simpler at the cost of runtime issues that could be detected by the compiler? I am not sure how likely that would be, but sounds like some debug session to find out what is going wrong. Just to realize that the API was called in the wrong way.

ryanmoran:
Changing the primary `DetectFunc` API to provide a compile-time assertion of invalid behavior feels like a big cost in API churn and complexity for a relatively small benefit of helping stack extension creators that don't fully grasp the buildpack specification. I think the runtime error message is clear and easily resolved. I'm really hesitant to mangle the API in such a way to prevent an invalid case that is likely very rare.

@paketo-buildpacks/tooling-maintainers thoughts?

robdimsdale:
I have a preference for minimizing breaking changes where possible - especially in this case where the breaking change (i.e. introducing generics) only improves the user experience a little bit.

I think we all agree that in principle, a compile-time failure is better than a run-time failure. It sounds like we disagree in principle on whether the cost of introducing generics (potential code bloat and complexity) is worth it to improve the UX in this way. In addition to the question of whether generics are the right choice in principle, we have a pragmatic question of whether introducing them now is worth making a breaking change to the packit v2 API.

Personally, I'm in favor of generics for this feature in principle, but I'd like to wait until we have other breaking changes to bundle them up and release a packit v3.

c0d1ngm0nk3y:
> Also considering the API, I think we can live with a single `Run` method that takes either a `BuildFunc` or a `GenerateFunc` as its second argument. We can achieve this with an interface type.

But isn't that breaking as well? If you have something like [this](https://github.com/paketo-buildpacks/packit/blob/3a413fdf9dc499785acb231baf71906eedfa75b4/run_test.go#L91).

```go
        build := func(packit.BuildContext) (packit.BuildResult, error) {
		buildCalled = true
		return packit.BuildResult{}, nil
	}

	detect := func(packit.DetectContext) (packit.DetectResult, error) {
		detectCalled = true
		return packit.DetectResult{}, nil
	}

	packit.Run(detect, build, packit.WithArgs(args), packit.WithExitHandler(exitHandler))
```

Then `build` will implement `BuildFunc` (old signature), but not `BuildOrGenerateFunc` (new signature). Or am I missing something?

run_test.go
```diff
@@ -20,8 +20,22 @@ func testRun(t *testing.T, context spec.G, it spec.S) {
 		tmpDir      string
 		cnbDir      string
 		exitHandler *fakes.ExitHandler
+
+		buildCalled  bool
+		detectCalled bool
 	)
 
+	var build packit.BuildFunc
```
c0d1ngm0nk3y: This line is actually needed because of the introduction of `_BuildOrGenerate`. So strictly speaking, this would be a breaking change, or?
@ryanmoran WDYT?

ryanmoran: You are right. I didn't run into this in my testing because we tend to do something like this:

```go
func Build() packit.BuildFunc {
	return func(context packit.BuildContext) (packit.BuildResult, error) { ... }
}
```

I guess I am less caught up in trying to make this API read nicely. I'd be fine with 2 Run funcs.

mhdawson:
I tried this out on our work in progress extension for ubi8 and was able to get it working and remove a lot of scafolding code that we needed beforehand.  Looking forward to this landing.

We are still working on unit tests so we don't have a set of integration tests to try, but the simple test of using the builder after the integration of this PR into packit and our extension validated that I could still build/run the application with ubi.

ryanmoran:
@c0d1ngm0nk3y, any progress on this? We'd love to get this into a good state soon.

c0d1ngm0nk3y:
> any progress on this? We'd love to get this into a good state soon.

I planned on looking into this again last Friday, but I got sick. Will probably look into this today or tomorrow...

c0d1ngm0nk3y:
@ryanmoran I have added a second `RunExtension` function to be more compatible. If that is the right balance between code reuse and readability, I will add an integration test. Or what else is missing? WDYT?

build_plan.go
```diff
@@ -19,6 +19,21 @@ type BuildPlan struct {
...
+type ExtensionBuildPlan struct {
+	// Provides is a list of BuildPlanProvisions that are provided by this
+	// extension.
+	Provides []BuildPlanProvision `toml:"provides"`
+
+	// Or is a list of additional BuildPlans that may be selected by the
+	// lifecycle
+	Or []ExtensionBuildPlan `toml:"or,omitempty"`
+}
+
```
ryanmoran:
```suggestion
```

I think this is no longer used, right?

ryanmoran (commented):
@c0d1ngm0nk3y this looks good to me, other than the dead code that I think we can remove. I also think we should be able to get to a diff that doesn't include any updates to the `sbom` package. Let me know if you are seeing issues with that and I can help.

generate.go
```diff
@@ -0,0 +1,159 @@
...
+
+	var info struct {
+		APIVersion string `toml:"api"`
+		Info       Info   `toml:"extension"`
+	}
+
+	extensionTOML := filepath.Join(cnbPath, "extension.toml")
+	_, err = toml.DecodeFile(extensionTOML, &info)
+	if err != nil {
+		config.exitHandler.Error(errors.Wrapf(err, "Could not parse '%s'", extensionTOML))
```
ryanmoran:
```suggestion
		config.exitHandler.Error(fmt.Errorf("Could not parse %q: %w", extensionTOML, err))
```

Could we used stdlib here now that it supports error wrapping? The `github.com/pkg/errors` package is archived.

ryanmoran (changes requested):

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/onsi/gomega v1.27.1
 	github.com/pelletier/go-toml v1.9.5
+	github.com/pkg/errors v0.9.1
```
ryanmoran: I think this will disappear if you run `go mod tidy`.

c0d1ngm0nk3y:
@ryanmoran Your comments have been addressed, I think or is there anything we missed?

ryanmoran (approved):

PR #460, opened 2023-02-23
Fix confusion in boolean parameter order in tests

phil9909:
## Summary

The test helpers `AssertEncoderAgainstGoldenImageSnapshot` and `AssertEncoderAgainstGoldenSnapshot` took two boolean parameters:  `updateSnapshot bool` and `json bool`. Some tests seem to have confused their order making the test act "weird".

To prevent it from happening again we introduced an enum `Type` with the values `TypePlain` and `TypeJson`.

## Use Cases

Improve tests

## Checklist

### DIFF
--- a/sbom/internal/formats/syft2/encoder_test.go
+++ b/sbom/internal/formats/syft2/encoder_test.go
@@ -23,8 +23,8 @@ func TestDirectoryEncoder(t *testing.T) {
 	testutils.AssertEncoderAgainstGoldenSnapshot(t,
 		Format(),
 		testutils.DirectoryInput(t),
-		true,
 		*updateJson,
+		testutils.TypeJson,
 	)
 }
 
@@ -34,8 +34,8 @@ func TestImageEncoder(t *testing.T) {
 		Format(),
 		testutils.ImageInput(t, testImage, testutils.FromSnapshot()),
 		testImage,
-		true,
 		*updateJson,
+		testutils.TypeJson,
 	)
 }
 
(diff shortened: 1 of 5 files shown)


robdimsdale (approved):

robdimsdale:
Thanks for this. I think this captures the intent more clearly.

When you say:

> Some tests seem to have confused their order making the test act "weird".

Can you point to the tests that mixed up the order? I didn't see any changes like that in this diff? Unless you mean the tests are in other buildpacks, not packit?

sbom/internal/formats/syft2/encoder_test.go
```diff
@@ -23,8 +23,8 @@ func TestDirectoryEncoder(t *testing.T) {
 	testutils.AssertEncoderAgainstGoldenSnapshot(t,
 		Format(),
 		testutils.DirectoryInput(t),
-		true,
 		*updateJson,
+		testutils.TypeJson,
```
phil9909, line 27: > Can you point to the tests that mixed up the order? I didn't see any changes like that in this diff?

E.g. here

you can see it was `true, *updateJson` now it is `*updateJson, testutils.TypeJson`

This caused the test to:
- compare the file not as a json unless you provide the `--update-json` flag, which is supposed to update the golden files and default to false
- always update the golden files

There are multiple similar cases.

> Unless you mean the tests are in other buildpacks, not packit?

This should not be the case, as it is in the `internal` sub-package (otherwise this would be a breaking change).

robdimsdale: Oh, yeah, of course. Thanks! It was so subtle that I missed it despite actively looking for it.

ForestEckhardt (approved):

PR #486, opened 2023-04-14
Add support for buildpack API v0.9

### LINKED ISSUE #431: Buildpack API 0.9: Remove Shell-Specific Logic
## Describe the Enhancement

Implement this [CNB RFC](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md) to remove support for shell-specific logic when a buildpack uses API 0.9.

The primary change here is that the structure of the `processes` section in the `launch.toml` has changed.

## Motivation

Having support for the latest API versions is important to support not only our own project usecases, but those of our users as well.

modulo11:
## Summary

This PR adds compatible support for buildpack API v0.9 fixing https://github.com/paketo-buildpacks/packit/issues/431. 

## Use Cases

It currently supports three "modes" to stay compatible also with older API versions:

- API < v0.9: No changes required, direct and indirect processes are possible
- API >= v0.9: If only the BP API version is bumped, the provided [`Process`](https://github.com/sap-contributions/packit/blob/c97e53b8120b11bcc311a7208ca75d6ad2a475b7/process.go#L18-L39) struct is translated into a [`DirectProcess`](https://github.com/sap-contributions/packit/blob/c97e53b8120b11bcc311a7208ca75d6ad2a475b7/process.go#L76-L94) using [`ToDirectProcess`](https://github.com/sap-contributions/packit/blob/c97e53b8120b11bcc311a7208ca75d6ad2a475b7/process.go#L54-L68). If it was used as an indirect process, it will be [translated](https://github.com/sap-contributions/packit/blob/c97e53b8120b11bcc311a7208ca75d6ad2a475b7/process.go#L56-L58) by prepending `bash` to the original command. I would put this up for discussion if it is the right thing to do. Maybe failing or supporting either the old or the new struct is a better way.
- API >= v0.9: Using [node-start](https://github.com/paketo-buildpacks/node-start/blob/main/build.go) as an example, the new facilities could be used like this:

```go
func Build(applicationFinder ApplicationFinder, logger scribe.Emitter, reloader Reloader) packit.BuildFunc {
...
		originalProcess := packit.DirectProcess{
			Type:    "web",
			Command: []string{"node"},
			Args:    []string{file},
			Default: true,
		}
...
		logger.LaunchDirectProcesses(processes)
		return packit.BuildResult{
				DirectProcesses: processes,
			},
		}, nil
	}
}
```

The separate [`LaunchProcess`](https://github.com/sap-contributions/packit/blob/c97e53b8120b11bcc311a7208ca75d6ad2a475b7/process.go#L5-L10) interface is a bit ugly but it was the easiest way to enable compatibility with the [`emitter`](https://github.com/sap-contributions/packit/blob/c97e53b8120b11bcc311a7208ca75d6ad2a475b7/scribe/emitter.go#L133), but I'm open for alternatives as well. One neat way could be to use generics, but the project is currently using go 1.16. 

## Checklist

### DIFF
--- a/build.go
+++ b/build.go
@@ -144,6 +144,7 @@ func Build(f BuildFunc, options ...Option) {
 	apiV05, _ := semver.NewVersion("0.5")
 	apiV06, _ := semver.NewVersion("0.6")
 	apiV08, _ := semver.NewVersion("0.8")
+	apiV09, _ := semver.NewVersion("0.9")
 	apiVersion, err := semver.NewVersion(buildpackInfo.APIVersion)
 	if err != nil {
 		config.exitHandler.Error(err)
@@ -284,13 +285,27 @@ func Build(f BuildFunc, options ...Option) {
 		}
 
 		var launch struct {
-			Processes []Process  `toml:"processes"`
-			Slices    []Slice    `toml:"slices"`
-			Labels    []label    `toml:"labels"`
-			BOM       []BOMEntry `toml:"bom"`
+			Processes       []Process       `toml:"processes,omitempty"`
+			DirectProcesses []DirectProcess `toml:"processes,omitempty"`
+			Slices          []Slice         `toml:"slices"`
+			Labels          []label         `toml:"labels"`
+			BOM             []BOMEntry      `toml:"bom"`
 		}
 
-		launch.Processes = result.Launch.Processes
+		if apiVersion.LessThan(apiV09) {
+			if result.Launch.DirectProcesses != nil {
+				config.exitHandler.Error(errors.New("direct processes can only be used with Buildpack API v0.9 or higher"))
+				return
+			}
+			launch.Processes = result.Launch.Processes
+		} else {
+			if result.Launch.Processes != nil {
+				config.exitHandler.Error(errors.New("non direct processes can only be used with Buildpack API v0.8 or lower"))
+				return
+			}
+			launch.DirectProcesses = result.Launch.DirectProcesses
+		}
+
 		if apiVersion.LessThan(apiV06) {
 			for _, process := range launch.Processes {
 				if process.Default {
--- a/process.go
+++ b/process.go
@@ -1,14 +1,5 @@
 package packit
 
-import "strings"
-
-type LaunchProcesses interface {
-	GetType() string
-	GetCommand() []string
-	GetArgs() []string
-	GetDefault() bool
-}
-
 // Process represents a process to be run during the launch phase as described
 // in the specification lower than v0.9:
 // https://github.com/buildpacks/spec/blob/main/buildpack.md#launch. The
@@ -38,19 +29,6 @@ type Process struct {
 	WorkingDirectory string `toml:"working-directory,omitempty"`
 }
 
-func (p Process) GetType() string {
-	return p.Type
-}
-func (p Process) GetCommand() []string {
-	return strings.Split(p.Command, " ")
-}
-func (p Process) GetArgs() []string {
-	return p.Args
-}
-func (p Process) GetDefault() bool {
-	return p.Default
-}
-
 // DirectProcess represents a process to be run during the launch phase as described
 // in the specification higher or equal than v0.9:
 // https://github.com/buildpacks/spec/blob/main/buildpack.md#launch. The
@@ -75,17 +53,4 @@ type DirectProcess struct {
 	// directory other than the application directory. This can either be an
 	// absolute path or one relative to the default application directory.
 	WorkingDirectory string `toml:"working-directory,omitempty"`
-}
-
-func (p DirectProcess) GetType() string {
-	return p.Type
-}
-func (p DirectProcess) GetCommand() []string {
-	return p.Command
-}
-func (p DirectProcess) GetArgs() []string {
-	return p.Args
-}
-func (p DirectProcess) GetDefault() bool {
-	return p.Default
 }
--- a/process.go
+++ b/process.go
@@ -1,7 +1,16 @@
 package packit
 
+import "strings"
+
+type LaunchProcesses interface {
+	GetType() string
+	GetCommand() []string
+	GetArgs() []string
+	GetDefault() bool
+}
+
 // Process represents a process to be run during the launch phase as described
-// in the specification:
+// in the specification lower than v0.9:
 // https://github.com/buildpacks/spec/blob/main/buildpack.md#launch. The
 // fields of the process are describe in the specification of the launch.toml
 // file:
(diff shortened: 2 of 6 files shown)


sophiewigmore:
Hey @modulo11 I will try to take a look at this PR this week. Sorry for the delay

process.go
```diff
@@ -28,3 +37,71 @@ type Process struct {
...
+	return p.Args
+}
+func (p Process) GetDefault() bool {
+	return p.Default
+}
+
+func (p Process) ToDirectProcess() DirectProcess {
+	command := []string{}
+	if !p.Direct {
+		command = append(command, "bash")
```
sophiewigmore: nit: I think we need to append `"-c"` as well to this

build.go
```diff
@@ -284,13 +285,26 @@ func Build(f BuildFunc, options ...Option) {
...
+			BOM             []BOMEntry      `toml:"bom"`
+		}
+
+		if apiVersion.LessThan(apiV09) {
+			launch.Processes = result.Launch.Processes
+		} else {
+			launch.DirectProcesses = []DirectProcess{}
+			if result.Launch.Processes != nil {
+				for _, process := range result.Launch.Processes {
+					launch.DirectProcesses = append(launch.DirectProcesses, process.ToDirectProcess())
```
sophiewigmore: I'm trying to decide what my stance is on converting the any non-direct processes into direct ones, versus. issuing an error when the process isn't well-formed for the API version.

On one hand, doing the conversion you have here makes the buildpack-author experience simpler by abstracting away the specifics. It also makes the upgrade path to API v0.9 easier. I think I prefer this.

On the other hand, it could be confusing to a buildpack author that a process is being converted under the hood. It might be more straightforward to just enforce the correct process structure.

@paketo-buildpacks/tooling-maintainers any opinions?

scribe/emitter.go
```diff
@@ -111,15 +111,35 @@ Entries:
...
+		launchProcesses = append(launchProcesses, process)
+	}
+	e.launch(launchProcesses, processEnvs...)
+}
+
+// LaunchDirectProcesses take a list of processes and a map of process specific
+// enivronment varables and prints out a formatted table including the type
+// name, whether or not it is a default process, the command, arguments, and
+// any process specific environment variables.
+func (e Emitter) LaunchDirectProcesses(processes []packit.DirectProcess, processEnvs ...map[string]packit.Environment) {
```
sophiewigmore, line 166: Also related to the `process.ToDirectProcess()` conversation above, I'm wondering what the log emitter experience would be in the case where an indirect process is used with API 0.9.
Which log function would we expect a buildpack author to use in their buildpack? `emitter.LaunchProcesses`?

build.go
modulo11: My understanding is that API v0.9 pushes the decision to use a shell back to the individual builpack instead of deciding it centrally in the `lifecycle` (via indirect process). If a buildpack maintainer consciously decides to update to v0.9 maybe it's better to consciously decide on what and how processes should be started, instead of again deciding it centrally in `packit`. On the other hand side, `packit` is more high-level library from which uses might expect that level of convenience of auto-conversion.

This conversion could also be added in a separate PR on demand.

robdimsdale: This is a great conversation. I can see the pros and cons of both approaches - packit being dumb and relying on the buildpack author to do the right thing, or packit trying to be opinionated and save the buildpack author from extra work.

My perspective is that it packit should stay fairly dumb on this point. I think it would be really confusing to debug a situation where a buildpack claimed it was compatible with API `v0.9`, and provided indirect processes, and yet still magically worked because packit automatically converted the indirect processes to direct processes.

As a buildpack author, if I upgrade to API `v0.9` I would expect to make the necessary changes to my buildpack to ensure it is compatible with that updated API version. I would be surprised if it just worked because Packit intervened, and I'd be worried that packit might do the wrong thing with my command in some other situation.

I think it's better to fail fast and inform the user/buildpack author that the process type must be direct if the API is `v0.9` or higher.

It looks similar to the situation where we fail fast if a process is marked as `Default` for API `v0.5` or lower. In this situation, Packit could be more intelligent and try to remove the `Default` label - especially if there is only one process - but I think it's more useful to fail and tell the user that they're trying to do something that contravenes the spec.

sophiewigmore: Thanks both for weighing in, it is an interesting problem :). I think Rob's point that we have the previous experience of failing fast in the `v0.5` API default process example he provided is good enough reason to move forward without auto process conversion.
I totally see the value in doing the conversion, but it's probably more important to keep the packit experience consistent.

modulo11: Thanks for the input and sorry for the delay. I'll try to get the PR ready this week.

robdimsdale: No rush!

sophiewigmore (dismissed):

sophiewigmore:
@modulo11 this looks great, thanks for updating the PR. When you have a chance, can you rebase it with the latest `v2` branch changes?
Also giving time for @robdimsdale to take another look if he wants to before we merge it in.

robdimsdale:
@modulo11 There's a refactor that I'd like to do, but I don't have permissions to push to your fork. Essentially I want to move the `GetXYZ` methods off the public API and into an internal interface within `scribe/emitter.go`. This change keeps the same functionality while also maintaining a minimal public API. These methods are only used within `scribe/emitter.go`, so this refactor continues to work. And I don't think it's particularly useful to add these methods to the public API of `Process.go`.

Could you either: A) give me permissions to write to your fork. Or B) apply [this commit](https://github.com/paketo-buildpacks/packit/commit/1ad99b1419b606ac2a019a872a074709a66e99b3).

modulo11:
> @modulo11 There's a refactor that I'd like to do, but I don't have permissions to push to your fork. Essentially I want to move the `GetXYZ` methods off the public API and into an internal interface within `scribe/emitter.go`. This change keeps the same functionality while also maintaining a minimal public API. These methods are only used within `scribe/emitter.go`, so this refactor continues to work. And I don't think it's particularly useful to add these methods to the public API of `Process.go`.
> 
> Could you either: A) give me permissions to write to your fork. Or B) apply [this commit](https://github.com/paketo-buildpacks/packit/commit/1ad99b1419b606ac2a019a872a074709a66e99b3).

That makes sense. I rebased the branch and picked your commit.

sophiewigmore:
@modulo11 can you add a basic test case for `emitter.LaunchDirectProcesses` ?
Otherwise this looks good to me!

sophiewigmore (approved):

robdimsdale (approved):

PR #563, opened 2024-04-04
Allows users to set a dependency mirror

TisVictress:
Implements [dependency-mirrors RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0060-dependency-mirrors.md)

A new release of each buildpack will need to be cut with the updated version of packit containing this change.

Reason for this change:
While there is already support for dependency mappings, this adds support for dependency mirrors. The difference is the user would not need to specify a mapping for each individual dependency. Dependency mirrors are more efficient because the user can set one mirror for all or multiple dependencies.

The mirror can be set via an environment variable or a binding of type `dependency-mirror`. In addition, multiple mirror URLs can be set, a default mirror with extra specific hostname mirrors. For example, if a user wants to download all dependencies from the same place except for dependencies that originated in github (see below). The default mirror can contain a placeholder, `{originalHost}`, to support mirrors that host many different repositories, which in some cases include the original hostname in the path.  If only specific hostname mirrors are set, the appropriate dependencies will be download via those mirrors while everything else will be downloaded through the original URLs in the buildpack's metadata.

Example mirrors via environment variables:
```
BP_DEPENDENCY_MIRROR              https://mirror.example.org/{originalHost}
BP_DEPENDENCY_MIRROR_GITHUB_COM   https://mirror.example.org/public-github
BP_DEPENDENCY_MIRROR_NODEJS_ORG   https://mirror.example.org/node-dist
```
Example of mirrors via binding
```
/platform
    └── bindings
        └── dependency-mirror
            ├── default                https://mirror.example.org/{originalHost}
            ├── github.com             https://mirror.example.org/public-github
            ├── nodejs.org             https://mirror.example.org/node-dist
            └── type                   dependency-mirror
```

If a dependency mapping and a dependency mirror is set, an error will be generated. Dependency mappings and mirrors cannot be used simultaneously.

If a binding is used to set dependency mirrors, there can only be one binding of type `dependency-mirror`, containing multiple `Entries` for each mirror URL.

If both methods, environment variable(s) and binding, are used to set a dependency mirror, the mirror(s) from the environment variable(s) will be chosen, as specified in the above RFC.

### DIFF
--- a/postal/internal/dependency_mirror.go
+++ b/postal/internal/dependency_mirror.go
@@ -1,0 +1,138 @@
+package internal
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"strings"
+)
+
+type DependencyMirrorResolver struct {
+	bindingResolver BindingResolver
+}
+
+func NewDependencyMirrorResolver(bindingResolver BindingResolver) DependencyMirrorResolver {
+	return DependencyMirrorResolver{
+		bindingResolver: bindingResolver,
+	}
+}
+
+func formatAndVerifyMirror(mirror, uri string) (string, error) {
+	mirrorURL, err := url.Parse(mirror)
+	if err != nil {
+		return "", err
+	}
+
+	uriURL, err := url.Parse(uri)
+	if err != nil {
+		return "", err
+	}
+
+	if mirrorURL.Scheme != "https" && mirrorURL.Scheme != "file" {
+		return "", fmt.Errorf("invalid mirror scheme")
+	}
+
+	mirrorURL.Path = strings.Replace(mirrorURL.Path, "{originalHost}", uriURL.Host+uriURL.Path, 1)
+	return mirrorURL.String(), nil
+}
+
+func (d DependencyMirrorResolver) FindDependencyMirror(uri, platformDir string) (string, error) {
+	mirror, err := d.findMirrorFromEnv(uri)
+	if err != nil {
+		return "", err
+	}
+
+	if mirror != "" {
+		return formatAndVerifyMirror(mirror, uri)
+	}
+
+	mirror, err = d.findMirrorFromBinding(uri, platformDir)
+	if err != nil {
+		return "", err
+	}
+
+	if mirror != "" {
+		return formatAndVerifyMirror(mirror, uri)
+	}
+
+	return "", nil
+}
+
+func (d DependencyMirrorResolver) findMirrorFromEnv(uri string) (string, error) {
+	const DefaultMirror = "BP_DEPENDENCY_MIRROR"
+	const NonDefaultMirrorPrefix = "BP_DEPENDENCY_MIRROR_"
+	mirrors := make(map[string]string)
+	environmentVariables := os.Environ()
+	for _, ev := range environmentVariables {
+		pair := strings.SplitN(ev, "=", 2)
+		key := pair[0]
+		value := pair[1]
+
+		if !strings.Contains(key, DefaultMirror) {
+			continue
+		}
+
+		if key == DefaultMirror {
+			mirrors["default"] = value
+			continue
+		}
+
+		// convert key
+		hostname := strings.SplitN(key, NonDefaultMirrorPrefix, 2)[1]
+		hostname = strings.ReplaceAll(strings.ReplaceAll(hostname, "__", "-"), "_", ".")
+		hostname = strings.ToLower(hostname)
+		mirrors[hostname] = value
+
+		if !strings.Contains(uri, hostname) {
+			continue
+		}
+
+		return value, nil
+	}
+
+	if mirrorUri, ok := mirrors["default"]; ok {
+		return mirrorUri, nil
+	}
+
+	return "", nil
+}
+
+func (d DependencyMirrorResolver) findMirrorFromBinding(uri, platformDir string) (string, error) {
+	bindings, err := d.bindingResolver.Resolve("dependency-mirror", "", platformDir)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve 'dependency-mirror' binding: %w", err)
+	}
+
+	if len(bindings) > 1 {
+		return "", fmt.Errorf("cannot have multiple bindings of type 'dependency-mirror'")
+	}
+
+	if len(bindings) == 0 {
+		return "", nil
+	}
+
+	mirror := ""
+	entries := bindings[0].Entries
+	for hostname, entry := range entries {
+		if hostname == "default" {
+			mirror, err = entry.ReadString()
+			if err != nil {
+				return "", err
+			}
+			continue
+		}
+
+		if !strings.Contains(uri, hostname) {
+			continue
+		}
+
+		mirror, err = entry.ReadString()
+		if err != nil {
+			return "", err
+		}
+
+		return mirror, nil
+	}
+
+	return mirror, nil
+}
--- a/postal/internal/dependency_mirror_test.go
+++ b/postal/internal/dependency_mirror_test.go
@@ -1,0 +1,285 @@
+package internal_test
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/paketo-buildpacks/packit/v2/postal/internal"
+	"github.com/paketo-buildpacks/packit/v2/postal/internal/fakes"
+	"github.com/paketo-buildpacks/packit/v2/servicebindings"
+	"github.com/sclevine/spec"
+
+	. "github.com/onsi/gomega"
+)
+
+func testDependencyMirror(t *testing.T, context spec.G, it spec.S) {
+	var (
+		Expect          = NewWithT(t).Expect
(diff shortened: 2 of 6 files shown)


TisVictress:
@sophiewigmore I've updated the description. Let me know if you have any questions.

postal/service.go
```diff
@@ -237,15 +254,28 @@ func (s Service) Deliver(dependency Dependency, cnbPath, layerPath, platformPath
 		dependencyChecksum = fmt.Sprintf("sha256:%s", dependency.SHA256)
 	}
 
+	dependencyMirrorURI, err := s.mirrorResolver.FindDependencyMirror(dependency.URI, platformPath)
```
sophiewigmore, line 260: nit - might want to update the function description comment to mention the dependency mirror

postal/service.go
```diff
@@ -237,15 +254,28 @@ func (s Service) Deliver(dependency Dependency, cnbPath, layerPath, platformPath
...
+		return fmt.Errorf("failure checking for dependency mirror: %s", err)
+	}
+
 	dependencyMappingURI, err := s.mappingResolver.FindDependencyMapping(dependencyChecksum, platformPath)
 	if err != nil {
 		return fmt.Errorf("failure checking for dependency mappings: %s", err)
 	}
 
+	if dependencyMirrorURI != "" && dependencyMappingURI != "" {
+		return fmt.Errorf("dependency mappings should not exist when using a dependency mirror")
```
sophiewigmore: what's your thinking in having this error out, over doing something like choosing a mirror to have precedence over a mapping?

sophiewigmore: Regardless we should do whatever they're doing in libpak so the experience is consistent

postal/internal/dependency_mirror.go
```diff
@@ -0,0 +1,143 @@
...
+	if mirror != "" {
+		return formatAndVerifyMirror(mirror, uri)
+	}
+
+	return "", nil
+}
+
+func (d DependencyMirrorResolver) findMirrorFromEnv(uri string) (string, error) {
+	const DefaultMirror = "BP_DEPENDENCY_MIRROR"
+	const DefaultMirrorPrefix = "BP_DEPENDENCY_MIRROR_"
```
sophiewigmore: for clarity, wouldn't this be the non-default mirror prefix?
```suggestion
	const NonDefaultMirrorPrefix = "BP_DEPENDENCY_MIRROR_"
```

postal/internal/dependency_mirror.go
```diff
@@ -0,0 +1,143 @@
...
+	mirrors := make(map[string]string)
+	environmentVariables := os.Environ()
+	for _, ev := range environmentVariables {
+		idx := strings.Index(ev, "=")
+		if idx < 0 {
+			continue
+		}
+
+		key := ev[:idx]
+		value := ev[idx+1:]
```
sophiewigmore: I think you can simplify this with something like:
```
pair := strings.SplitN(ev, "=", 2)
key := pair[0]
value := pair[1]
```

postal/internal/dependency_mirror.go
```diff
@@ -0,0 +1,143 @@
...
+		if key == DefaultMirror {
+			mirrors["default"] = value
+			continue
+		}
+
+		// convert key
+		hostnameIdx := len(DefaultMirrorPrefix)
+		hostname := key[hostnameIdx:]
+		hostname = strings.ReplaceAll(hostname, "_", ".")
+		hostname = strings.ReplaceAll(hostname, "__", "-")
```
sophiewigmore: This logic doesn't work perfectly, might need to add a test for it.

For example, if the input is `BP_DEPENDENCY_MIRROR_TESTING_123__ABC`, the expected output is `testing.123-abc`, but this code actually outputs `testing.123..abc`

I think these lines need to get reversed, so double underscores are converted first

postal/internal/dependency_mirror.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		if key == DefaultMirror {
+			mirrors["default"] = value
+			continue
+		}
+
+		// convert key
+		hostnameIdx := len(DefaultMirrorPrefix)
+		hostname := key[hostnameIdx:]
```
sophiewigmore: if you want to simplify this, you could do `hostname := strings.SplitN(key, DefaultMirrorPrefix, 2)[1]`

postal/internal/dependency_mirror_test.go
```diff
@@ -0,0 +1,213 @@
...
+		bindingResolver *fakes.BindingResolver
+		err             error
+	)
+
+	context("FindDependencyMirror", func() {
+		context("via binding", func() {
+			it.Before(func() {
+				tmpDir, err = os.MkdirTemp("", "dependency-mirror")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(os.WriteFile(filepath.Join(tmpDir, "default"), []byte("https://mirror.example.org/{originalHost}"), os.ModePerm))
```
sophiewigmore, line 30: Can you add a couple tests for the case where there's no default provided?

postal/internal/dependency_mirror_test.go
```diff
@@ -0,0 +1,213 @@
...
+					Expect(err).ToNot(HaveOccurred())
+					Expect(bindingResolver.ResolveCall.Receives.Typ).To(Equal("dependency-mirror"))
+					Expect(bindingResolver.ResolveCall.Receives.Provider).To(BeEmpty())
+					Expect(bindingResolver.ResolveCall.Receives.PlatformDir).To(Equal("some-platform-dir"))
+					Expect(boundDependency).To(Equal("https://mirror.example.org/public-github"))
+				})
+			})
+		})
+
+		context("via environment variables", func() {
```
sophiewigmore, line 126: same as above, can we add an example or two when there is no default provided?

sophiewigmore: also maybe a test when an env var and binding are both provided?

sophiewigmore:
Do we need to add anything to deal with the username/password parts of the URI? and what about the file scheme (`https` or `file`). Can the `postal.Deliver` method deal with both of these things already? Might need tests for these scenarios too if we need to change anything

postal/service.go
TisVictress: Hmmm...I think I might have made this up 🙃 It's not in the RFC and I dont see it in libpak so I'll remove it

TisVictress: This check also exists in [libpack](https://github.com/bitgully/libpak/blob/main/dependency_cache.go#L257). It doesn't error out but it defaults to the binding. I'll make this change.

TisVictress:
> Do we need to add anything to deal with the username/password parts of the URI?

If a username and password are included in the url, it's captured by `url.Parse` which is a go-builtin function.

TisVictress:
> Can the `postal.Deliver` method deal with both of these things already?

@sophiewigmore  I'll double check and get back to you.

TisVictress:
@sophiewigmore `cargo.Transport` [handles both](https://github.com/paketo-buildpacks/packit/blob/v2/cargo/transport.go#L19) `https` and `file`.

I did some adhoc/integration testing with the nginx buildpack, inserting `user:pass@` before the beginning of the uri and after the uri scheme. For example,
```
https://user:pass@artifacts.paketo.io/nginx/nginx_1.24.0_linux_x64_jammy_fa1dfe46.tgz
```
Usernames and passwords are handled by `cargo.Transport` implicitly because it does not tamper with the uri before [passing it directly to the http request](https://github.com/paketo-buildpacks/packit/blob/v2/cargo/transport.go#L28).  (The username and password are put in the `Authorization` [header](https://serverfault.com/questions/371907/can-you-pass-user-pass-for-http-basic-authentication-in-url-parameters) of the request)


I've updated the PR to address all the above suggestions.

postal/service.go
```diff
@@ -228,22 +245,37 @@ func stringSliceElementCount(slice []string, str string) int {
 // location of the CNBPath is given so that dependencies that may be included
 // in a buildpack when packaged for offline consumption can be retrieved. If
 // there is a dependency mapping for the specified dependency, Deliver will use
-// the given dependency mapping URI to fetch the dependency. The dependency is
-// validated against the checksum value provided on the Dependency and will
-// error if there are inconsistencies in the fetched result.
+// the given dependency mapping URI to fetch the dependency. If there is a
+// dependency mirror for the specified dependency, Deliver will use the mirror
+// URI to fetch the dependency. The dependency is validated against the checksum
```
sophiewigmore:
```suggestion
// URI to fetch the dependency. If both a dependency mapping and mirror are BOTH present, the mapping will take precedence over the mirror. The dependency is validated against the checksum
```

postal/service.go
sophiewigmore: Bump on this one - I see the lines commented out around this. Can you get rid of that since we've decided against the logger?

sophiewigmore (approved):

PR #569, opened 2024-04-24
Fixes mirror bug when originalHost is excluded

TisVictress:
Also adds tests to makes sure that url path is complete

### DIFF
--- a/postal/internal/dependency_mirror_test.go
+++ b/postal/internal/dependency_mirror_test.go
@@ -51,12 +51,12 @@ func testDependencyMirror(t *testing.T, context spec.G, it spec.S) {
 
 			context("given a default mirror binding", func() {
 				it("finds a matching dependency mirror in the platform bindings if there is one", func() {
-					boundDependency, err := resolver.FindDependencyMirror("some-uri", "some-platform-dir")
+					boundDependency, err := resolver.FindDependencyMirror("https://some-uri.com/dep.tgz", "some-platform-dir")
 					Expect(err).ToNot(HaveOccurred())
 					Expect(bindingResolver.ResolveCall.Receives.Typ).To(Equal("dependency-mirror"))
 					Expect(bindingResolver.ResolveCall.Receives.Provider).To(BeEmpty())
 					Expect(bindingResolver.ResolveCall.Receives.PlatformDir).To(Equal("some-platform-dir"))
-					Expect(boundDependency).To(Equal("https://mirror.example.org/some-uri"))
+					Expect(boundDependency).To(Equal("https://mirror.example.org/some-uri.com/dep.tgz"))
 				})
 			})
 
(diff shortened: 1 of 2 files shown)


postal/internal/dependency_mirror_test.go
```diff
@@ -213,14 +213,34 @@ func testDependencyMirror(t *testing.T, context spec.G, it spec.S) {
...
 				Expect(err).ToNot(HaveOccurred())
-				Expect(boundDependency).NotTo(Equal("https://mirror.example.org/some-uri"))
-				Expect(boundDependency).To(Equal("https://mirror.other-example.org/some-uri"))
+				Expect(boundDependency).NotTo(Equal("https://mirror.example.org/some-uri.com/dep.tgz"))
+				Expect(boundDependency).To(Equal("https://mirror.other-example.org/some-uri.com/dep.tgz"))
 
 			})
 		})
 
+		context("without originalHost placeholder", func() {
```
sophiewigmore, line 224: what does this mean?

sophiewigmore: Can you elaborate on what the bug that you're solving is?

TisVictress: The user can use `{originalHost}` as a placeholder to include the original url path when specifying a mirror. For example, if the original url is `https://artifacts.paketo.io/nginx_1.24.0_linux_x64_jammy_fa1dfe46.tgz` and the mirror is `https://user:pass@local-mirror.example.com/buildpacks-dependencies/{originalHost}`, then this will translate to `https://user:pass@local-mirror.example.com/buildpacks-dependencies/artifacts.paketo.io/nginx_1.24.0_linux_x64_jammy_fa1dfe46.tgz`. But if the placeholder is excluded, then the mirror translates to `https://user:pass@local-mirror.example.com/buildpacks-dependencies/nginx_1.24.0_linux_x64_jammy_fa1dfe46.tgz`

TisVictress:
@sophiewigmore In the initial change, the mirror url was not properly constructed and did not include the actual dependency archive. For instance, with the following:
original: `https://artifacts.paketo.io/nginx_1.24.0_linux_x64_jammy_fa1dfe46.tgz`
mirror: `https://user:pass@local-mirror.example.com/buildpacks-dependencies`
...it would not append the actual tgz file to be downloaded. The result instead was `https://user:pass@local-mirror.example.com/buildpacks-dependencies/artifacts.paketo.io`
This also happened with special hostnames:
original: `https://github.com/nginx/nginx_1.24.0_linux_x64_jammy_fa1dfe46.tgz`
mirror: `https://mirror.example.org/public-github`
...the result would incorrectly be `https://mirror.example.org/public-github` when it should be `https://mirror.example.org/public-github/nginx/nginx_1.24.0_linux_x64_jammy_fa1dfe46.tgz`

The initial change only worked correctly if the `{originalHost}` placeholder was used. This is due to line [35 of dependency_mirror.go](https://github.com/paketo-buildpacks/packit/pull/569/files#diff-222ddc4d38993468881693ce72a1893d05091f53ce0aa89b4e425118c1d889daR35). The path needed to be moved outside of the substitution.

I also updated the tests because the test urls did not include an appropriate scheme and path

sophiewigmore:
This is helpful info! thank you

sophiewigmore (approved):

PR #580, opened 2024-06-17
Standard dep metadata

### LINKED ISSUE #557: Implement RFC 0059: Standardized Dependency Metadata
We should implement [Paketo RFC 0059](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0059-standard-dependency-metadata-format.md) to standardize dependency metadata.

We should do this in a backwards-compatible manner, so buildpacks can move over to the new structure as they please, without breaking

This work will be important for unblocking ARM64/multi-architecture buildpack support, by enabling dependency metadata for architecture specific dependencies.

thitch97:
## Summary

This serves as the initial implementation of RFC 0059, wherein `packit` adopts the accepted Standardized Dependency Metadata format. As part of this implementation changes were made in the `cargo`, `postal`, `scribe`, and `sbom` packages to accommodate the new metadata fields. For backwards compatibility, the new fields exist alongside the deprecated fields instead of replacing them entirely. 

Unresolved Issues:
- The standardized format defines a `PURLs` field (i.e. a set of PURLs) instead of just a single PURL for each dependency. This makes sense in theory, but the `syft` library we use to generate SBoMs only [allows one PURL per package declaration](https://github.com/paketo-buildpacks/packit/blob/e366827e9d3aab77a0b9abc9ceb8829c7eeb7faf/sbom/internal/formats/syft2/model/package.go#L29). To get around this, I thought it appropriate to create a `syft` [catalog with a package for each unique PURL](https://github.com/paketo-buildpacks/packit/blob/e366827e9d3aab77a0b9abc9ceb8829c7eeb7faf/sbom/sbom.go#L100) as opposed to just a single package. However, I have yet to find a suitable method of testing this approach.

- Though the RFC states: _"Note: Both the distros and licenses fields are optional, however if they are given then the non-optional components of them must be set."_, I'm not sure what an appropriate place is to enforce that rule (perhaps the parsing logic in `cargo` and `postal`).

Resolves #557 

## Checklist

### DIFF
--- a/cargo/config.go
+++ b/cargo/config.go
@@ -39,6 +39,11 @@ type ConfigBuildpackLicense struct {
 	URI  string `toml:"uri"  json:"uri"`
 }
 
+type ConfigBuildpackDistro struct {
+	Name    string `toml:"name" json:"name"`
+	Version string `toml:"version"  json:"version"`
+}
+
 type ConfigMetadata struct {
 	IncludeFiles          []string                             `toml:"include-files"              json:"include-files,omitempty"`
 	PrePackage            string                               `toml:"pre-package"                json:"pre-package,omitempty"`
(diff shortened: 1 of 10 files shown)


ForestEckhardt:
> * The standardized format defines a `PURLs` field (i.e. a set of PURLs) instead of just a single PURL for each dependency. This makes sense in theory, but the `syft` library we use to generate SBoMs only [allows one PURL per package declaration](https://github.com/paketo-buildpacks/packit/blob/e366827e9d3aab77a0b9abc9ceb8829c7eeb7faf/sbom/internal/formats/syft2/model/package.go#L29). To get around this, I thought it appropriate to create a `syft` [catalog with a package for each unique PURL](https://github.com/paketo-buildpacks/packit/blob/e366827e9d3aab77a0b9abc9ceb8829c7eeb7faf/sbom/sbom.go#L100) as opposed to just a single package. However, I have yet to find a suitable method of testing this approach.

I think that your idea makes sense unfortunately I am not sure that I have a much better idea when it comes to how to test this.

> * Though the RFC states: _"Note: Both the distros and licenses fields are optional, however if they are given then the non-optional components of them must be set."_, I'm not sure what an appropriate place is to enforce that rule (perhaps the parsing logic in `cargo` and `postal`).

I don't think that there needs to be enforcement of this it would just mean that any `buildpack.toml` that did not conform would not be compliant and therefore there are no guarantees that it will be compatible with the Paketo Buildpacks ecosystem. I think if critical information is missing then the build will fail and if it is not critical then maybe there will be some malformed logging or SBOM generation and I think that is sufficient for me.

PR #622, opened 2024-12-20
Enable architecture specific dependencies

### LINKED ISSUE #618: Support architecture for dependencies
## Describe the Enhancement

When resolving a buildpack dependency, the architecture should be taken into account. That would allow the same version to be present for different architectures (e.g. `arm` and `amd`). This would allow `arm` support for buildpacks using `packit`.

## Possible Solution

The [Resolve](https://github.com/paketo-buildpacks/packit/blob/f9409a4d1f917000d8a506f2f6b2991193bc2e2e/postal/service.go#L106) function should take the architecture in consideration when looking for a dependency. Similar to the [Resolve](https://github.com/paketo-buildpacks/libpak/blob/ec910fcece3a160ce8ad2806830e679f119cc823/buildpack.go#L494) function in `libpak`. But the solution should be less "hacky" (e.g. taking the architecture from a property rather the `purl`)

## Motivation

https://github.com/paketo-buildpacks/nodejs/issues/995

In order to get `arm` support for the `nodejs` meta buildpack, there must be a way to filter dependencies on their architecture.

c0d1ngm0nk3y:
fixes https://github.com/paketo-buildpacks/packit/issues/618

## Summary

When looking for an dependency, the architecture is taken into account. The default architecture (if not provided by a dependency is "amd64" to ensure backwards compatibility. The architecture to look for is taken from the system or an env variable `BP_ARCH` to enable unit testing. 

## Use Cases

Dependencies for several architectures (e.g. "arm" and "amd64")

## Checklist

### DIFF
--- a/postal/buildpack.go
+++ b/postal/buildpack.go
@@ -74,6 +74,9 @@ type Dependency struct {
 	// StripComponents behaves like the --strip-components flag on tar command
 	// removing the first n levels from the final decompression destination.
 	StripComponents int `toml:"strip-components"`
+
+	// ARCH is the architecture of this dependency
+	Arch string `toml:"arch"`
 }
 
 func parseBuildpack(path, name string) ([]Dependency, string, error) {
--- a/postal/service.go
+++ b/postal/service.go
@@ -4,8 +4,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"sort"
 	"strings"
 	"time"
@@ -52,15 +54,17 @@ type ErrNoDeps struct {
 	id                string
 	version           string
 	stack             string
+	arch              string
 	supportedVersions []string
 }
 
 // Error implements the error.Error interface
 func (e *ErrNoDeps) Error() string {
-	return fmt.Sprintf("failed to satisfy %q dependency version constraint %q: no compatible versions on %q stack. Supported versions are: [%s]",
+	return fmt.Sprintf("failed to satisfy %q dependency version constraint %q: no compatible versions on %q stack with architecture %q. Supported versions are: [%s]",
 		e.id,
 		e.version,
 		e.stack,
+		e.arch,
 		strings.Join(e.supportedVersions, ", "),
 	)
 }
@@ -147,6 +151,10 @@ func (s Service) Resolve(path, id, version, stack string) (Dependency, error) {
 			continue
 		}
 
+		if !isCorrectArch(dependency) {
+			continue
+		}
+
 		sVersion, err := semver.NewVersion(dependency.Version)
 		if err != nil {
 			return Dependency{}, err
@@ -160,7 +168,7 @@ func (s Service) Resolve(path, id, version, stack string) (Dependency, error) {
 	}
 
 	if len(compatibleVersions) == 0 {
-		return Dependency{}, &ErrNoDeps{id, version, stack, supportedVersions}
+		return Dependency{}, &ErrNoDeps{id, version, stack, archFromSystem(), supportedVersions}
 	}
 
 	stacksForVersion := map[string][]string{}
@@ -218,6 +226,25 @@ func (s Service) Resolve(path, id, version, stack string) (Dependency, error) {
 	})
 
 	return compatibleVersions[0], nil
+}
+
+func isCorrectArch(dep Dependency) bool {
+	systemArch := archFromSystem()
+
+	if systemArch == "amd64" && dep.Arch == "" {
+		return true
+	}
+
+	return systemArch == dep.Arch
+}
+
+func archFromSystem() string {
+	archFromEnv, ok := os.LookupEnv("BP_ARCH")
+	if !ok {
+		archFromEnv = runtime.GOARCH
+	}
+
+	return archFromEnv
 }
 
 func stringSliceContains(slice []string, str string) bool {
--- a/postal/service_test.go
+++ b/postal/service_test.go
@@ -23,6 +23,7 @@ import (
 	//nolint Ignore SA1019, usage of deprecated package within a deprecated test case
 	"github.com/paketo-buildpacks/packit/v2/paketosbom"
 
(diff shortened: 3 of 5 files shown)


c0d1ngm0nk3y:
@paketo-buildpacks/tooling-maintainers Could someone have a look? That is a prerequisite for supporting multiarch in paketo-buildpacks/nodejs.

ForestEckhardt (approved):

ForestEckhardt:
Could you please rebase this PR so I can merge it in!

postal/buildpack.go
```diff
@@ -74,6 +74,9 @@ type Dependency struct {
 	// StripComponents behaves like the --strip-components flag on tar command
 	// removing the first n levels from the final decompression destination.
 	StripComponents int `toml:"strip-components"`
+
+	// ARCH is the architecture of this dependency
+	Arch string `toml:"arch"`
```
pacostas, line 79: There are several platforms, for example, pack cli has plenty of them when it comes to build and image https://github.com/buildpacks/pack/blob/5cca9c51a61cf1bb8b0b6ba479bbbf10a65f0377/internal/target/platform.go#L31-L47 For nodejs we have a subset of them https://github.com/nodejs/node/blob/main/BUILDING.md#platform-list which I think it would be good to start with. So my suggestion on this one would be to also add another variable called `OS` or `operating-system` which will indicate the operating system of the dependency

postal/service.go
```diff
@@ -52,15 +54,17 @@ type ErrNoDeps struct {
...
 
 // Error implements the error.Error interface
 func (e *ErrNoDeps) Error() string {
-	return fmt.Sprintf("failed to satisfy %q dependency version constraint %q: no compatible versions on %q stack. Supported versions are: [%s]",
+	return fmt.Sprintf("failed to satisfy %q dependency version constraint %q: no compatible versions on %q stack with architecture %q. Supported versions are: [%s]",
 		e.id,
 		e.version,
 		e.stack,
+		e.arch,
 		strings.Join(e.supportedVersions, ", "),
```
pacostas, line 68: Same on this one we should add the os attribute

postal/service_test.go
```diff
@@ -115,6 +116,12 @@ strip-components = 1
 		service = postal.NewService(transport).
 			WithDependencyMappingResolver(mappingResolver).
 			WithDependencyMirrorResolver(mirrorResolver)
+
+		Expect(os.Setenv("BP_ARCH", "amd64")).To(Succeed())
```
pacostas, line 120: On this one, instead of having BP_ARCH and BP_OS, we could rename the `BP_ARCH` to something like BP_PLATFORM or BP_TARGET, which will include both architecture and operating system, similar to what pack cli does with the targets.

postal/service.go
```diff
@@ -220,6 +228,25 @@ func (s Service) Resolve(path, id, version, stack string) (Dependency, error) {
...
+	systemArch := archFromSystem()
+
+	if systemArch == "amd64" && dep.Arch == "" {
+		return true
+	}
+
+	return systemArch == dep.Arch
+}
+
+func archFromSystem() string {
```
pacostas, line 241: In terms of clean code, one function should only do one thing, I would remove from this function, the two things that is doing 1. looking if the environment variable exists 2. fetching the system architecture. Changing that function to fetching only the architecture does not make sense as there is already a function for that. Therefore, I would suggest to move this logic out in to the Parent function (`Resolve`) as I think this is totally normal as there is already logic for validating other things like (`dependency version` and `stacks`). Althouh, if you decide to fetch the platform, then it would be good to wrap these two lines in a function called `getSystemPlatform` as it will be more than one line (one for fetching the `GOARCH` and one for fetching the `GOOS`

pacostas: I think we dont want to default into the system that the user is using as pack currently fallbacks to linux/amd64, so I would choose this platform as default, in order to avoid any errors from macos users that currently building applications

pacostas (changes requested):

postal/service.go
c0d1ngm0nk3y: > I think we dont want to default into the system

I might be completely wrong, but that is not what is done. It looks up the architecture of the runtime, meaning the buildpack that actually called `getDependency()`. So no other architecture could even work. The `BP_ARCH` is only needed for unit tests.

I tried to keep everything as similar to [`libak`](https://github.com/paketo-buildpacks/libpak/blob/aa7107c9e10714dce0adc2a13d85aa26cd739253/buildpack.go#L566) as possible.

postal/buildpack.go
c0d1ngm0nk3y: I am no expert on the whole multiarch thing, but `operating-system` would be wrong, wouldn't it? The operating system is for all dependencies `linux` - it is the same run image after all. We "only" have to differ dependencies for the architecture of the image. So I thing anything other than `Arch` would be quite confusing, don't you think?

c0d1ngm0nk3y:
> Could you please rebase this PR so I can merge it in!

I will update, but wait a bit with the merge so that @pacostas can respond to the comments.

postal/buildpack.go
pacostas: I thought that we might need to support generating images for macos, but currently there is no such use case so makes sense to not over-complicate it. LGTM

postal/service.go
pacostas: If we dont use the os, this comment does not apply.

postal/service.go
pacostas: I see, so seems that it is not breaking the default behavior. LGTM

postal/service_test.go
pacostas: If we dont use the os, this comment does not apply.

pacostas:
> > Could you please rebase this PR so I can merge it in!
> 
> I will update, but wait a bit with the merge so that @pacostas can respond to the comments.

LGTM

c0d1ngm0nk3y:
@ForestEckhardt Would you merge it?

PR #657, opened 2025-06-26
Update service binding implementation to match libcnb

### LINKED ISSUE #658: Service binding resolution order discrepancy between packit and libcnb
## Expected Behavior
I expect the resolution order to be the same between packit and libcnb because the paketo language family buildpacks use a mix of component buildpacks that leverage both those libraries. 

## Current Behavior
libcnb reads bindings from SERVICE_BINDING_ROOT first, then VCAP_SERVICES whereas packit does the reverse

## Possible Solution
Packit should be updated to match libcnb because VCAP_SERVICES should not take precedence over explicitly defined SERVICE_BINDING_ROOT

## Steps to Reproduce
https://github.com/buildpacks/libcnb/blob/274a25f73782e8af70dfb13909b870520ac23dd1/platform.go#L248-L262

## Motivations
I am using cnb on cloud foundry and am seeing inconsistent behavior between buildpacks

### LINKED ISSUE #659: VCAP_SERVICES parsing does not respect explicit type provided in credential
## Expected Behavior
I expected the parsing to mimic the k8s service binding parsing where the type is inferred from the contents of a type file.

## Current Behavior
The type is inferred from the label in the cf binding credential 

## Possible Solution
VCAP_SERVICE parsing should respect the type if it is in the binding credential, otherwise it should fall back to the existing behavior (using the label)

## Steps to Reproduce
create a user provided service binding with a field called type and see that it is not respected when bound in cloud foundry

## Motivations
user-provided service bindings are always inferred to be type user-provided which doesn't match any known binding type in cnb.

tomkennedy513:
## Summary
The ordering of when VCAP_SERVICES was parsed was inconsistent between buildpack libraries resulting in unexpected behavior so I updated it to match libcnb
Also added logic to read type from a field called type in the binding when parsing VCAP_SERVICES.

I also removed the caching of service bindings because the caching is only beneficial if resolve is called multiple times from the same buildpack. Due to the fact that few buildpacks even support multiple binding types and that the binding resolution does not have massive overhead (there are typically not many bindings present) I didn't see a need to keep the complexity. 

## Use Cases
This change will result in consistent behavior between the two buildpack libraries. By parsing the type file we will also create consistency with k8s service bindings and allows a service broker to set an explicit type. it also allows for user provided services to define the type they want because using the label will always result in type 'user-provided'.

fixes #658 
fixes #659 

## Checklist

### DIFF
--- a/servicebindings/resolver.go
+++ b/servicebindings/resolver.go
@@ -33,10 +33,9 @@ type Binding struct {
 //
 // It also supports backwards compatibility with the legacy service binding spec:
 // https://github.com/buildpacks/spec/blob/main/extensions/bindings.md
-type Resolver struct {
-	bindingRoot string
-	bindings    []Binding
-}
+//
+// It also supports reading bindings from VCAP_SERVICES
+type Resolver struct{}
 
 // NewResolver returns a new service binding resolver.
 func NewResolver() *Resolver {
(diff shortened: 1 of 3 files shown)


servicebindings/resolver.go
```diff
@@ -50,19 +49,17 @@ func NewResolver() *Resolver {
...
-	if newRoot := bindingRoot(platformDir); r.bindingRoot != newRoot {
-		r.bindingRoot = newRoot
-		bindings, err := loadBindings(r.bindingRoot)
-		if err != nil {
-			return nil, fmt.Errorf("failed to load bindings from '%s': %w", r.bindingRoot, err)
-		}
-		r.bindings = bindings
+	bindings, err := loadBindings(platformDir)
+	if err != nil {
+		// return nil, fmt.Errorf("failed to load bindings: %w", err)
```
robdimsdale: Can we either uncomment this line or delete it please?

robdimsdale (dismissed):

servicebindings/resolver.go
tomkennedy513: yes will delete

robdimsdale (approved):

PR #662, opened 2025-06-29
Add OS and Distros to dependency metadata

### LINKED ISSUE #650: Architecture validation is too restrictive
Commit be25401c29f2c5e05f10e97fdcdfcd900ebba745 enabled architecture specific dependencies however the current implementation breaks python package handling (maybe others).

## Expected Behavior

Installing noarch python package should work on all supported architectures.

## Current Behavior

Since 2.17.0, noarch packages can only be installed on amd64. However arm64 is also a valid target which does not work anymore.

## Possible Solution

Two possible solutions:
- Make (and document) a list of officially supported architectures and add arm64 to it.
- If dependency.Arch is empty, don't check the system architecture.

## Steps to Reproduce

1. Clone the pip buildpack
2. Update go.mod using `go mod tidy`
3. Build the buildpack `./scripts/package.sh --version 99`
4. Create a dummy python project containing a `requirements.txt` file
5. Use pack to build the project using the custom pip buildpack `pack build my-test-project --builder paketobuildpacks/builder-jammy-base --path . --verbose --buildpack paketo-buildpacks/cpython --buildpack ~/buildpacks/pip/build/buildpackage.cnb --buildpack  paketo-buildpacks/python-start`

## Motivations

This new check breaks at least the pip buildpack on non amd64 architectures.
It also breaks my current work on the python buildpacks consolidation.

jericop:
## Summary

This change adds `OS` and `Distros` to buildpack.toml dependency metadata in accordance to [this RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0059-standard-dependency-metadata-format.md) and updates the existing unit tests to verify they work as expected. It also adds comments above unit test lines that fail when not running in github actions. This happens because there are directory paths in the data and they will almost certainly differ when running locally.

resolves https://github.com/paketo-buildpacks/packit/issues/650

This change is required before a PR to jam in order to prevent buildpacks from dropping OS and Arch metadata during packaging.

## Checklist

### DIFF
--- a/cargo/config.go
+++ b/cargo/config.go
@@ -24,8 +24,13 @@ type ConfigStack struct {
 }
 
 type ConfigTarget struct {
-	OS     string   `toml:"os"     json:"os,omitempty"`
-	Arch   string   `toml:"arch"   json:"arch,omitempty"`
+	OS   string `toml:"os"       json:"os,omitempty"`
+	Arch string `toml:"arch"     json:"arch,omitempty"`
+}
+
+type ConfigDistro struct {
+	Name    string `toml:"name" json:"name"`
+	Version string `toml:"version"  json:"version"`
 }
 
 type ConfigBuildpack struct {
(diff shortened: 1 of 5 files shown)


dmikusa:
@jericop Is this consistent with the RFC on metadata format? I want to make sure that we're not diverging from what's in the RFC. If it does, and there's good reason, we should update the RFC.

https://github.com/paketo-buildpacks/rfcs/blob/main/text/0059-standard-dependency-metadata-format.md

Variant is in particular what I'm wondering about.

jericop:
> @jericop Is this consistent with the RFC on metadata format? I want to make sure that we're not diverging from what's in the RFC. If it does, and there's good reason, we should update the RFC.
> 
> https://github.com/paketo-buildpacks/rfcs/blob/main/text/0059-standard-dependency-metadata-format.md
> 
> Variant is in particular what I'm wondering about.

Good catch. Variant is mentioned in two cnb RFCs which is why I added it, but when I look at the environment variable set by the lifecycle in a test build locally I don't see env `CNB_TARGET_VARIANT` being set. I will remove variant and update the title. I will also add distros and ensure dependency metadata parity with libpak.

jericop:
This should now be ready for review. Please let me know if you see anything else that needs to be changed.

dmikusa (approved):

PR #671, opened 2025-07-27
Only enforce strict arch checking when dependency specifies arch

jericop:
## Summary

This PR does the following:
* Disables strict dependency checking based on architecture
* Only set BP_ARCH for unit tests related to dependencies with architecture
* Removes the architecture from the error message because it can be confusing for noarch dependencies such as pip and yarn
* Fixes typos in a few comments

this is related to https://github.com/paketo-buildpacks/packit/commit/be25401c29f2c5e05f10e97fdcdfcd900ebba745
resolves https://github.com/paketo-buildpacks/packit/issues/650
relates to https://github.com/paketo-buildpacks/packit/issues/618
relates to https://github.com/paketo-buildpacks/nodejs/issues/995

## Use Cases

The [0059-standard-dependency-metadata-format](https://github.com/paketo-buildpacks/rfcs/blob/e601a1b0bbf21b54da05588df876fb1d989e2497/text/0059-standard-dependency-metadata-format.md?plain=1#L54-L56) RFC says that when `os` and `arch` are not given it should be assumed that the dependency is compatible with any architecture. This will be a requirement for some buildpacks such as [pip](https://github.com/paketo-buildpacks/pip/blob/58c86600835adab848d144c155c8e1f7002c4dbf/buildpack.toml#L25) and [yarn](https://github.com/paketo-buildpacks/yarn/blob/8b4b8e00f2168f22d4e7b0ddc29e7510f59d5a04/buildpack.toml#L29), which work on any architecture.

This PR makes it so that when dependencies specify `arch` they can only be used on systems with the specified `arch` or if env `BP_ARCH` is set to match. This means strict checking is only enabled when dependencies specify `arch`.

## Checklist

* [x] I have viewed, signed, and submitted the Contributor License Agreement.
* [x] I have linked issue(s) that this PR should close using keywords or the Github UI (See [docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue))
* [x] I have added an integration test, if necessary.
* [x] I have reviewed the [styleguide](https://github.com/paketo-buildpacks/community/blob/main/STYLEGUIDE.md) for guidance on my code quality.
* [x] I'm happy with the commit history on this PR (I have rebased/squashed as needed).

PR #687, opened 2025-09-09
Fetching dependencies based on the CNB_TARGET_ARCH and CNB_TARGET_OS

pacostas:
## Summary

This PR filters the dependencies from the buildpack.toml based on the CNB_TARGET_ARCH and on the CNB_TARGET_OS env variables. In case these variables are not available, it will fall back to the host architecture and os

In addition, this PR removes filtering dependencies based on the the BP_ARCH env variable.

## Checklist

### DIFF
--- a/postal/buildpack.go
+++ b/postal/buildpack.go
@@ -115,9 +115,9 @@ func stacksInclude(stacks []string, stack string) bool {
 
 func supportsPlatform(targetOs, targetArch string, dependency Dependency) bool {
 
+	// Avoid strict checking in case of dependency does not specify OS/Arch
 	if dependency.OS == "" && dependency.Arch == "" {
-		dependency.OS = "linux"
-		dependency.Arch = "amd64"
+		return true
 	}
 
 	if targetOs != dependency.OS || targetArch != dependency.Arch {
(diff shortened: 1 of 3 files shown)


postal/buildpack.go
```diff
@@ -112,3 +112,17 @@ func stacksInclude(stacks []string, stack string) bool {
 	}
 	return false
 }
+
+func platformInclude(targetOs, targetArch string, dependency Dependency) bool {
```
jericop: I would suggest renaming `platformInclude` to something like `supportsPlatform` for clarity.

postal/buildpack.go
```diff
@@ -112,3 +112,17 @@ func stacksInclude(stacks []string, stack string) bool {
...
+
+	if dependency.OS == "" && dependency.Arch == "" {
```
jericop, line 119: In the spirit of https://github.com/paketo-buildpacks/packit/pull/671, this should just return true because we don't want to do any os/arch checking when the dependency does not specify any.

pacostas: I see, this will allow to build apps on any arch without having to specify the os and arch on the dependencies.

jericop (approved):

PR #723, opened 2025-12-26
fix: add x-shellscript to NopArchive

sgaist:
## Summary

Some scripts are now detected as text/x-shellscript rather than only text.

Without this fix, the archive handling fails.

## Use Cases

In it's current state, this breaks the Miniconda buildpack which uses
a script to install miniconda.

## Checklist


sgaist:
~I haven't added an integration test yet as I am unsure what would be the best way to do it.~
I tested the patch locally with the miniconda buildpack using a replace directive pointing to my fixed version of packit.

sgaist:
With the linting error fixes, it should also unblock #714 and #717

fs/copy.go
```diff
@@ -45,13 +45,17 @@ func copyFile(source, destination string) error {
...
-	defer sourceFile.Close()
+	defer func() {
+		_ = sourceFile.Close()
+	}()
 
 	destinationFile, err := os.Create(destination)
 	if err != nil {
 		return err
 	}
-	defer destinationFile.Close()
```
pacostas, line 54: I think here it would be nice to throw an error if the destinatoin file can not be closed.

internal/file_writer.go
```diff
@@ -16,7 +16,9 @@ func (fw FileWriter) Write(path string, reader io.Reader) error {
 	if err != nil {
 		return err
 	}
-	defer file.Close()
```
pacostas, line 19: I think also here it wold be nice to throw an error in case the file.close has thrown an error

internal/toml_writer.go
```diff
@@ -17,7 +17,9 @@ func (tw TOMLWriter) Write(path string, value interface{}) error {
 	if err != nil {
 		return err
 	}
-	defer file.Close()
```
pacostas, line 20: I think throwing an error in case the file.close has failed, it would be nice.

vacation/archive.go
pacostas, line 39: Can you also please add the `"text/x-shellscript"` on the comment here?

pacostas (changes requested):

vacation/archive.go
sgaist: Done !

fs/copy.go
sgaist: I refactored all similar deferred functions to return their error as part of the function flows. That way the behaviour should be consistent for the library.
However I just realized that you might have meant something else by throwing e.g. using `panic`.
Which one is it ?

internal/file_writer.go
sgaist: See https://github.com/paketo-buildpacks/packit/pull/723#discussion_r2671941731

internal/toml_writer.go
sgaist: See https://github.com/paketo-buildpacks/packit/pull/723#discussion_r2671941731

fs/copy.go
jericop: Your changes address what @pacostas is looking for, which is handling the close error.
I like the approach of joining all the errors. Looks clean. I confirmed that the MatchError (used in the unit test) supports joined errors as well as wrapped errors, so we are good there.
https://pkg.go.dev/github.com/onsi/gomega#MatchError

pacostas (approved):

pacostas:
@sgaist Thank you soo much for your contribution!! Good job!

jericop (approved):

